PR #42, opened 2022-12-14
Implement contract deposits/withdraws and token contract

jamesondh:
## Motivation

The three main feature additions in this PR are:
- A new seda-token NEAR contract
- Token deposits and withdraws for node operators on the mainchain contract
- Integration tests to test depositing and withdrawing (in `contracts/tests/`)

### SEDA Token

This contract uses the NEP-141 standard. An image and more metadata will have to be added in a future PR.

### Mainchain contract

New public functions added:
- deposit (through ft_transfer_call)
- withdraw
- withdraw_all

Nodes track the balances of node owner who have deposited into the mainchain contract. For simplification purposes, the `Node` list has been changed from a `LookupMap` to an `UnorderedMap` to make lookup easier using the account ID, remove the need to store `num_nodes`, and remove the functions related to changing ownership and unregistering. A new `HumanReadableNode` struct has been created for `get_node()` and `get_nodes()` to attach the account ID to the struct. 

## Testing

Unit tests are limited in this PR since staking requires calls to a token contract (to fetch token balances and transfer tokens), so integration tests are provided as a separate cargo project in `contracts/tests/`.

## Related PRs and Issues

note: unrelated tests are failing due to an out-of-scope issue and will be fixed in another PR


gluax (approved):

contracts/mainchain/src/consts.rs
```diff
@@ -0,0 +1,20 @@
...
+pub const PROMISE_CALL: u64 = 5_000_000_000_000;
+pub const GAS_FOR_FT_ON_TRANSFER: Gas = Gas(BASE_GAS + PROMISE_CALL);
+
+/// The amount of yocto NEAR the contract dedicates to guarantee that the
+/// "share" price never decreases. It's used during rounding errors for share ->
+/// amount conversions.
+/// The amount of yocto NEAR the contract dedicates to guarantee that the
+/// "share" price never decreases. It's used during rounding errors for share ->
+/// amount conversions.
+pub const STAKE_SHARE_PRICE_GUARANTEE_FUND: Balance = 1_000_000_000_000;
```
FranklinWaller, line 13: From what I understand this is the price per share? Those should be in SEDA and are not a constant (They can be changed by the DAO)

contracts/mainchain/src/staking.rs
```diff
@@ -0,0 +1,374 @@
...
+    ///
+    /// price = total_staked / total_shares
+    /// Price is fixed
+    /// (total_staked + amount) / (total_shares + num_shares) = total_staked /
+    /// total_shares (total_staked + amount) * total_shares = total_staked *
+    /// (total_shares + num_shares) amount * total_shares = total_staked *
+    /// num_shares num_shares = amount * total_shares / total_staked
+    pub(crate) fn num_shares_from_staked_amount_rounded_down(&self, amount: Balance) -> NumStakeShares {
+        assert!(self.total_staked_balance > 0, "The total staked balance can't be 0");
+        (U256::from(self.total_stake_shares) * U256::from(amount) / U256::from(self.total_staked_balance)).as_u128()
```
FranklinWaller, line 185: The amount of shares a staker should get back is not dependent on totals. The price per share is a setting from the DAO. So the calculation should be `amount / price_per_share` rounded down.

contracts/mainchain/src/consts.rs
jamesondh: No, this is a guarantee fund of a dust amount of SEDA token for rounding errors when converting from shares to token amount. Price per share is not explicitly stored on the contract and is derived from total staked balance / total shares, which changes when new tokens are minted to the contract.

contracts/mainchain/src/staking.rs
jamesondh: `num_shares_from_staked_amount_rounded_down()` is called in `stake()` to calculate the amount to shares to grant the user. For example, say there is 900 tokens already staked in the contract and that 1 token = 1 share for now (i.e. no rewards have been released yet). Bob deposits and stakes 100 tokens. The shares he gets is calculated as 900 total shares * 100 new tokens staked by Bob / 900 total staked tokens = 100 shares to Bob.

This logic is forked from NEAR's core staking-pool contract btw:
https://github.com/near/core-contracts/tree/master/staking-pool

mariocao (commented):
I was checking some of the contracts and I think we should add some references (if applies ofc). For example, the seda token contract is from the near documentation or examples repo.

PR #55, opened 2023-01-18
feat(runtime): Add ability to push events to the event queue from WASM

FranklinWaller:
## Motivation

Some tasks require to be executed by a different WASM binary, or in a different VM environment (Like DataRequests, which should have limited capabilities). These decisions on what to do when are decided by the WASM binary, but it needs to be able to trigger events to access these capabilities. This PR introduces trigger_event as a new PromiseAction to enable this.

## Explanation of Changes

Added a trigger_event to the PromiseAction's to enable event pushing to the event queue. I had to include the app actor inside the host actor to let it be able to send messages to the app actor.
At startup the app sends it's own address to the host actor so it can be used later.

## Testing

Run the node normally and trigger the command: `make run-build-all -- -c near cli hello` this will call the node first with the hello world and push an event to the queue (ChainTick). You can see the node executing the VM again with this event.

### DIFF
--- a/node/src/errors.rs
+++ b/node/src/errors.rs
@@ -14,6 +14,8 @@ pub enum NodeError {
     RuqliteError(#[from] rusqlite::Error),
     #[error("Chain Adapter Error: {0}")]
     ChainAdapterError(#[from] seda_chains::ChainAdapterError),
+    #[error("Missing app actor address in host adapter, was the node booted?")]
+    MissingAppActorAddress,
 }
 
 pub type Result<T, E = NodeError> = core::result::Result<T, E>;
--- a/node/src/host/trigger_event.rs
+++ b/node/src/host/trigger_event.rs
@@ -1,0 +1,29 @@
+//! Communication layer between App & Host adapter
+//! We send a message to trigger an event to the host actor
+//! which redirects the message to the app actor
+use actix::prelude::*;
+use seda_runtime::HostAdapter;
+use seda_runtime_sdk::events::Event;
+use serde::{Deserialize, Serialize};
+
+use crate::{event_queue_handler::AddEventToQueue, Host, NodeError::MissingAppActorAddress, Result};
+
+#[derive(Message, Serialize, Deserialize)]
+#[rtype(result = "Result<()>")]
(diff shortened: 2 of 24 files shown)


node/src/host/trigger_event.rs
```diff
@@ -0,0 +1,29 @@
+/// Communication layer between App & Host adapter
```
gluax, line 1: Is this supposed to be on the struct?

Or file level documentation? If it is file level the doc string should be //! instead of ///

node/src/errors.rs
```diff
@@ -14,6 +14,8 @@ pub enum NodeError {
     RuqliteError(#[from] rusqlite::Error),
     #[error("Chain Adapter Error: {0}")]
     ChainAdapterError(#[from] seda_chains::ChainAdapterError),
+    #[error("Missing app actor address in host adapter, was the node booted?")]
```
gluax, line 17: Is this error even possible to hit? This is in `Node` so the Node should be booted? Unless it's safety of someone using the library outside of our repo.

runtime/core/src/host_adapter.rs
```diff
@@ -2,7 +2,7 @@ use std::fmt::Display;
 
 use seda_chains::Client;
 use seda_config::{ChainConfigs, NodeConfig};
-use seda_runtime_sdk::Chain;
+use seda_runtime_sdk::{events::Event, Chain};
 
 #[async_trait::async_trait]
 pub trait HostAdapter: Send + Sync + Unpin + 'static {
```
gluax, line 8: I question the need for this to still be a trait. But even if we change that it should be in another PR.

FranklinWaller: Yeah agree, let's discuss this later

node/src/errors.rs
FranklinWaller: It probably won't hit, but since we are avoiding unwraps, and in the rare case it might happen, there's at least something for it

mariocao (approved):

gluax (approved):

PR #56, opened 2023-01-18
Improve CLI

gluax:
## Motivation

The existing CLI commands were outdated, so they needed to be updated to work.

## Explanation of Changes

Breaks down the CLI into a few different parts.

Best represented  by the image below:
image

The `sub-chain` command consists of the `call` and `view`  commands. They work the same way they did in the WASM CLI. However, after discussing with @mariocao, we decided this was only useful to developers. Hence, they only show in a debug build of `seda`. The ability to do `call` and `view` remains in the WASM binaries.

Debug mode:
image

Release mode(no sub-chain sub-command):
image

The `node` commands are the same as they were in the WASM CLI.
image

The `generate` command generates autocompletion for your shell environment.

The `run` command is the same as before.

The commands have been reorganized for cleanliness purposes.

**BUG FIXES**: There was a bug with how the `view` method on the `chain` trait worked before, where it couldn't always deserialize to a `String`. Also, we were turning it back into bytes in multiple places anyways. So for ease, `view` has been changed to return bytes.

**TODOS**:

- Consider moving `NodeResult` to somewhere WASM can also use it.
- Generate a man page for the `seda` binary.
- Clean up chain trait:
  - Most of the methods there are unused and unnecessary. We only need `call`, `view`, and any methods those rely on.
- Conditionally compile the usage of `clap` for the config module. It only needs `clap` if you use it in the CLI. 

## Testing

I manually tested that all the commands still work.

## Related PRs and Issues

None


gluax:
Draft PR because I forgot I still need to update the README and write the CLI section.

cli/src/cli/commands/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    .await?;
+    let _result = chain::send_tx(chain, client, &signed_txn).await?;
+
+    serde_json::to_writer_pretty(
+        std::io::stdout(),
+        &json!({
+                "status": "success"
+        }),
+    )
+    .expect("TODO");
```
FranklinWaller, line 52: Forgotten todo

cli/src/cli/commands/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    chain: Chain,
+    contract_id: &str,
+    method_name: &str,
+    args: String,
+    chains_config: &ChainConfigs,
+) -> crate::Result<()> {
+    let client = Client::new(&chain, chains_config)?;
+    let result = chain::view(chain, client, contract_id, method_name, args.into_bytes()).await?;
+    let value = serde_json::from_slice::<T>(&result).expect("TODO");
+    serde_json::to_writer_pretty(std::io::stdout(), &value).expect("TODO");
```
FranklinWaller, line 66: Forgotten todo

cli/src/cli/commands/node/get_node.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct GetNode {
+    #[arg(short, long)]
+    pub node_id:     u64,
+    #[arg(short, long)]
+    pub contract_id: Option<String>,
+}
+
+impl GetNode {
+    pub async fn handle(self, config: AppConfig, chains_config: PartialChainConfigs) -> Result<()> {
+        let chains_config = config.chains.to_config(chains_config)?;
```
FranklinWaller, line 19: Would it be better to have the parsing done in the upper level instead of letting every command parse the config by themselves?

gluax: I thought about that, but our clean code practices state that functions shouldn't be too large. Personally, if I put all this logic in the above match statements, it looks too large to me.

chains/src/lib.rs
```diff
@@ -110,28 +118,28 @@ pub mod chain {
...
-    ) -> Result<String> {
+    ) -> Result<Vec<u8>> {
         match chain {
             Chain::Another => AnotherChain::view(client.another(), contract_id, method_name, args).await,
             Chain::Near => NearChain::view(client.near(), contract_id, method_name, args).await,
         }
     }
 
-    async fn _get_node_owner(chain: Chain, client: Client, params: Params<'_>) -> Result<String> {
+    async fn _get_node_owner(chain: Chain, client: Client, params: Params<'_>) -> Result<Vec<u8>> {
```
mennatabuelnaga, line 128: _get_node_owner(), _get_node_socket_address(), _get_nodes(), _register_node(), _remove_node(), _set_node_socket_address(), _sign_tx() could be removed

chains/src/near_chain.rs
```diff
@@ -149,9 +148,9 @@ impl ChainAdapterTrait for NearChain {
                     _ => return Err(ChainAdapterError::CallChangeMethod(err.to_string())),
                 },
                 Ok(response) => {
-                    info!("response gotten after: {}s", delta);
+                    debug!("response gotten after: {}s", delta);
```
mennatabuelnaga, line 151: could be removed

chains/src/near_chain.rs
```diff
@@ -149,9 +148,9 @@ impl ChainAdapterTrait for NearChain {
...
 
-                    info!("response.status: {:#?}", response.status);
+                    debug!("response.status: {:#?}", response.status);
```
mennatabuelnaga, line 153: this too

mennatabuelnaga:
Looks great!
Could you please clean up the old cli comments in WASM?
Also, for some reason logging is no longer working from the node side whenever we call any of the cli commands.

cli/src/cli/commands/node/get_node.rs
FranklinWaller: True but DRY is I think more important (it causes more mistakes)

gluax: Oh, I need misunderstood your first comment. Yea, I could do that though we'd have to clone; that's not the worst since it's only a CLI command. But, the bigger issue and the reason why I have it in the individual handle sections is that those arguments may not be the same always. For instance, if we add a future command that doesn't require the `ChainsConfig` or `NodeConfig`. In fact, even now I have it set up so some only require the part of the `NodeConfig` they need for a cleaner UX.

chains/src/lib.rs
gluax: I was thinking of doing this in a future PR to keep this smaller. Similarly, we can remove most of the functions in the chain adapter trait. But, if desired, I could do that now. @mariocao @FranklinWaller, do you think I should do this now or in a future PR?

cli/src/cli/commands/node/get_node.rs
FranklinWaller: The individual handles is perfect, but I don't think a few handles not using some config while they were converted isn't a huge deal. It saves us some confusion later on

chains/src/lib.rs
FranklinWaller: I would say do it after this PR, this PR is pretty important for other tasks

FranklinWaller (approved):

mennatabuelnaga (approved):

PR #57, opened 2023-01-23
Feat: Data Request WASM

gluax:
## Motivation

To have a limited VM that only enables certain functionality.

## Explanation of Changes

Added a match pattern in the runtime that rejects the promise of any disallowed methods, but only during runtime.

Similarly, I made it so the SDK has a `full` feature flag. This way, users can also disable those functionalities at the built WASM binary level. 

## Testing

All current tests pass. I also added a new trial to ensure the promise was rejected when the VM was run with limited compatibility.

## Related PRs and Issues

N/A

### DIFF
--- a/runtime/sdk/src/promises/actions.rs
+++ b/runtime/sdk/src/promises/actions.rs
@@ -20,7 +20,11 @@ impl PromiseAction {
     pub fn is_limited_action(&self) -> bool {
         matches!(
             self,
-            Self::DatabaseGet(_) | Self::DatabaseSet(_) | Self::ChainCall(_) | Self::ChainView(_) | Self::TriggerEvent(_)
+            Self::DatabaseGet(_)
+                | Self::DatabaseSet(_)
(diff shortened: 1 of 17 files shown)


FranklinWaller (approved):

runtime/sdk/src/promises/actions.rs
```diff
@@ -13,6 +15,34 @@ pub enum PromiseAction {
...
+        matches!(
+            self,
+            Self::DatabaseGet(_)
+                | Self::DatabaseSet(_)
+                | Self::ChainCall(_)
+                | Self::ChainView(_)
+                | Self::TriggerEvent(_)
+        )
+    }
+}
```
mariocao, line 30: I think that it is better that by default everything is limited unless specified.

In this case we are explicitly defining what is limited and we should do the other way around.
The reason is that if we add a new action and we forgot to add it here, it will become unlimited by default.

PR #58, opened 2023-01-24
feat(p2p): Upgrade libp2p to v0.50.0

FranklinWaller:
## Motivation

Upgrade to the latest version of libp2p to keep up with the latest security updates.

## Explanation of Changes

Had to move some dependencies around to make it compatible with the latest version

### DIFF
--- a/p2p/src/libp2p/behaviour.rs
+++ b/p2p/src/libp2p/behaviour.rs
@@ -16,8 +16,8 @@ use libp2p::{
         ValidationMode,
     },
     identity::Keypair,
-    mdns::{Mdns, MdnsConfig, MdnsEvent},
-    NetworkBehaviour,
+    mdns::{self},
+    swarm::NetworkBehaviour,
 };
 
 use super::super::errors::Result;
@@ -30,7 +30,7 @@ pub struct SedaBehaviour {
     /// Message propagation
     pub gossipsub: Gossipsub,
     // TODO: change discovery mechanism
-    pub mdns:      Mdns,
+    pub mdns:      mdns::async_io::Behaviour,
 }
 
 impl SedaBehaviour {
@@ -55,7 +55,7 @@ impl SedaBehaviour {
         gossipsub.subscribe(&topic)?;
 
         Ok(Self {
-            mdns: Mdns::new(MdnsConfig::default()).await?,
+            mdns: mdns::async_io::Behaviour::new(mdns::Config::default())?,
             gossipsub,
         })
     }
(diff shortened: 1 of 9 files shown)


p2p/src/libp2p/behaviour.rs
```diff
@@ -16,21 +16,23 @@ use libp2p::{
...
-    mdns::{Mdns, MdnsConfig, MdnsEvent},
-    NetworkBehaviour,
+    mdns::{self},
+    swarm::NetworkBehaviour,
 };
 
 use super::super::errors::Result;
 use crate::P2PAdapterError;
 
+pub const GOSSIP_TOPIC: &str = "testnet";
```
gluax, line 26: I see this is also used in `mod.rs` I wonder if it's better to put it there, or even in a `consts.rs` file so it's easier to find?

p2p/src/libp2p/transport.rs
```diff
@@ -20,11 +20,13 @@ pub fn build_tcp_transport(key_pair: identity::Keypair) -> Result<transport::Box
     let noise_config = noise::NoiseConfig::xx(noise_keys).into_authenticated();
     let yamux_config = YamuxConfig::default();
 
-    Ok(TcpTransport::new(GenTcpConfig::default().nodelay(true))
+    let transport_config = Config::default().nodelay(true);
+    let transport = TcpTransport::new(transport_config)
         .upgrade(Version::V1)
         .authenticate(noise_config)
         .multiplex(yamux_config)
-        // TODO: use duration from p2p config
```
gluax, line 27: Can we keep note of this todo? Unless we are saying, we don't want that timeout configurable anymore. It doesn't have to be a comment in the codebase. We can put it as part of the issue in Jira as well.

p2p/src/libp2p/transport.rs
```diff
@@ -20,11 +20,13 @@ pub fn build_tcp_transport(key_pair: identity::Keypair) -> Result<transport::Box
     let noise_config = noise::NoiseConfig::xx(noise_keys).into_authenticated();
     let yamux_config = YamuxConfig::default();
 
-    Ok(TcpTransport::new(GenTcpConfig::default().nodelay(true))
+    let transport_config = Config::default().nodelay(true);
+    let transport = TcpTransport::new(transport_config)
```
gluax, line 24: I think it's cleaner to do `Ok(TcpTransport::new(transport_config)...)` here instead of binding to a variable and then returning that variable.

mariocao (approved):

gluax (approved):

PR #60, opened 2023-01-26
feat: node config in wasm

gluax:
## Motivation

It makes the node config accessible in WASM.

## Explanation of Changes

Rather than adding a new import, I leveraged the existing memory adapter to read and write. Upon creation, the memory adapter gets the `NodeConfig` and is then accessible as a global variable in WASM.

I made this global. Otherwise, it would have to deserialize the data every time. Now it's just deserialized once per VM.

## Testing

I added calling the config in a WASM test.

## Related PRs and Issues

None.

### DIFF
--- a/runtime/core/src/runtime.rs
+++ b/runtime/core/src/runtime.rs
@@ -108,6 +108,7 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
                         let stderr_pipe = Pipe::new();
 
                         let mut wasi_env = WasiState::new(&call_action.function_name)
+                            .env("WASM_NODE_CONFIG", serde_json::to_string(&self.node_config)?)
                             .args(call_action.args.clone())
                             .stdout(Box::new(stdout_pipe))
                             .stderr(Box::new(stderr_pipe))
@@ -137,6 +138,7 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
                             let mut stdout_buffer = String::new();
                             wasi_stdout.read_to_string(&mut stdout_buffer)?;
                             if !stdout_buffer.is_empty() {
+                                dbg!(&stdout_buffer);
                                 output.push(stdout_buffer);
                             }
                         }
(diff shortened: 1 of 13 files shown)


FranklinWaller:
I would rather leverage the ENV system (which is native in WASM) than using memory_read/memory_write for this: https://docs.rs/wasmer-wasi/0.17.1/wasmer_wasi/state/struct.WasiStateBuilder.html#method.env is a bit more standard

gluax:
@FranklinWaller I'll take a look at that!

gluax:
@FranklinWaller that does work, but then it won't allow us to have a global `CONFIG` type variable. At least not without doing an `init` call. Or just let the user have to grab the env variable and then convert it themselves to the config type.

runtime/sdk/src/wasm/config.rs
```diff
@@ -0,0 +1,14 @@
+use lazy_static::lazy_static;
+use seda_config::{NodeConfig, NodeConfigInner};
+
+use super::memory_read;
+
+fn config() -> NodeConfig {
+    let config_bytes = memory_read("*&_seda_node_config");
```
FranklinWaller, line 7: But isn't it just replacing this line with: https://doc.rust-lang.org/std/env/fn.var.html

gluax: Nope, since this function is run at compile time, the env var does not yet exist. But that's cause it's a global variable. We could just have the function public, and then it could be called. We could do that, but we should make it clear that they should only call this function once per instance otherwise, it's deserializing every call.

gluax: Actually nevermind it was an issue with serde that I have no resolved :)

mennatabuelnaga (approved):

runtime/core/src/runtime.rs
```diff
@@ -137,6 +138,7 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
                             let mut stdout_buffer = String::new();
                             wasi_stdout.read_to_string(&mut stdout_buffer)?;
                             if !stdout_buffer.is_empty() {
+                                dbg!(&stdout_buffer);
```
FranklinWaller, line 141: Remove the dbg's and you're good to go :)

FranklinWaller (approved):

mariocao (approved):

PR #61, opened 2023-01-26
[WIP] Feat/simple bridge

gluax:
## Motivation

A simple bridging mechanism in WASM.
Opening the DRAFT PR now to gather feedback on the implementation so far.

## Explanation of Changes

Creates a new consensus wasm directory with a cli structure slightly similar to our actual one.

Also added a CLI command, `seda node bridge` though it is mostly for debugging purposes as an easy way to trigger a bridge.

## Testing

For now, bridging from near to itself manually as a test.

## Related PRs and Issues

Blocked by #60.


wasm/consensus/src/tasks/bridge.rs
```diff
@@ -0,0 +1,68 @@
...
+        // it should be moved to the sdk
+        memory_write("bridge_deposit", self.deposit.to_le_bytes().to_vec());
+        chain_view(self.chain, self.contract_id, self.method_name, self.args.into_bytes())
+            .start()
+            .then(call_self("bridge", vec![]));
+    }
+}
+
+// Oh right wasmer doesn't have strings... or u128s so passing deposit as an arg
+// may not be good
```
FranklinWaller, line 31: A lot of chains use u128 with WASM, I know that NEAR for example uses json a lot to pass around values: https://docs.rs/near-sdk/2.0.0/near_sdk/json_types/struct.U128.html

wasm/consensus/src/main.rs
```diff
@@ -0,0 +1,17 @@
...
+mod tasks;
+
+#[derive(Debug, Parser)]
+struct Options {
+    #[command(subcommand)]
+    task: tasks::Task,
+}
+
+fn main() {
+    let options = Options::parse();
```
FranklinWaller, line 13: I'm fine with using clap, but are there any limitations to using this approach compared to calling exported functions (and using them as start_func?)

wasm/consensus/src/tasks/bridge.rs
gluax: The problem is `serde`, which both `serde_json` and `borsh` use, doesn't support `u128`. What they do is a work around of `to_string` and `parse`.  Doing bytes conversion is about `70x` faster on average than this approach. We might be able to pass an array of u8s or i32s(which all u8s should safely fit in) natively with wasm too to avoid the `memory_read` and `memory_write` methods.

gluax: Benchmarks for reference: 

image

wasm/consensus/src/main.rs
gluax: I did this for two reasons! 

One was not to have to do manual parsing of arguments ourselves. Granted, we could do an individual clap CLI per entry point. However, that feels less clean to me.

The second reason was to make it easier to call from the CLI leveraging the `call` RPC endpoint. Even if we only use this functionality for debug purposes, being able to test it from the CLI easily will make testing on our parts significantly easier. And this approach may also make eventual fuzzing easier, but that's a guess on my part.

gluax:
Still blocked till #60 is merged, but ignore the config changes in this PR, as those are updated and correct in the other PR.

wasm/consensus/src/tasks/bridge.rs
```diff
@@ -20,27 +20,27 @@ impl Bridge {
     pub fn handle(self) {
         // we have a method to auto convert to bytes in a trait in runtime.
         // it should be moved to the sdk
+        // TODO: SEDA-188 will make it so we can pass these instead of a vec of strings
+        // to .then()
         memory_write("bridge_deposit", self.deposit.to_le_bytes().to_vec());
         chain_view(self.chain, self.contract_id, self.method_name, self.args.into_bytes())
             .start()
-            .then(call_self("bridge", vec![]));
+            .then(call_self("step_1", vec![]));
```
FranklinWaller, line 28: I should've been more clear, I meant bridge_step_1 -> bridge_step_2, otherwise if we add another task the names collide

mennatabuelnaga (approved):

gluax:
No longer blocked :)

FranklinWaller (approved):

PR #62, opened 2023-01-30
feat(runtime): Add p2p broadcasting & receiving support

FranklinWaller:
## Motivation

Add a way for the runtime to send and receive p2p messages from inside the WASM runtime. 

## Explanation of Changes

It introduces two channels a message & command channel. The message channel is for getting a p2p message and dispatching that to our runtime. the command channel is for the runtime to dispatch a broadcast/unicast message. the p2p server listens to this channel and acts accordingly. 

## Testing

Run two servers and type a message inside the terminal when they are connected. The p2p message will be given to the wasm runtime and will send a message back

### DIFF
--- a/runtime/sdk/src/wasm/mod.rs
+++ b/runtime/sdk/src/wasm/mod.rs
@@ -8,6 +8,7 @@ mod execution;
 mod http;
 mod log;
 mod memory;
+mod p2p;
 mod promise;
 mod raw;
 
@@ -21,4 +22,5 @@ pub use execution::*;
 pub use http::*;
 pub use log::*;
 pub use memory::*;
+pub use p2p::*;
 pub use promise::*;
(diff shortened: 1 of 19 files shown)


runtime/sdk/src/wasm/mod.rs
```diff
@@ -19,4 +20,5 @@ pub use execution::*;
 pub use http::*;
 pub use log::*;
 pub use memory::*;
+pub use p2p::*;
```
gluax, line 23: Should p2p be allowed only in full?

gluax (commented):
Could we swap to `tokio::sync::mpsc`? It's still async methods.

The rationale being `futures::mpsc` is based on the standard library one, which is mal-performant and poorly implemented. Well, stdlib just rewrote their entire mpsc in Rust 1.67.0. But, I'm not sure the `futures` crate has yet to reflect those changes since that version of Rust was recently released.

gluax (approved):

PR #63, opened 2023-02-01
feat(contracts): implement get current slot and epoch

jamesondh:
## Motivation

Similarly to Ethereum 2.0, SEDA protocol is based on the concepts of slots and epochs. This PR introduces two functions on the NEAR mainchain contract to calculate both.

## Explanation of Changes

Added two public-facing functions:
- `get_current_slot()`
- `get_current_epoch()`

Both function are non-state modifying, calculated based on the NEAR block's height. This borrows two constants from the Ethereum PoS spec:
- NEAR_BLOCKS_PER_SEDA_SLOT to 10 (which averages 12 seconds per slot with NEAR blocks every 1.2 seconds)
- SLOTS_PER_EPOCH to 32 slots

## Testing

`slot_test.rs` includes tests for checking the slot and epoch while passing a mock block height. First it is set to block 0, then incremented to block 10 to check if the slot has increased (similar for epochs).

## Related PRs and Issues

n/a

### DIFF
--- a/contracts/src/epoch.rs
+++ b/contracts/src/epoch.rs
@@ -1,13 +1,13 @@
 use near_sdk::{env, near_bindgen};
 
-use crate::{slot::SECONDS_PER_SLOT, MainchainContract, MainchainContractExt};
+use crate::{slot::NEAR_BLOCKS_PER_SEDA_SLOT, MainchainContract, MainchainContractExt};
 
-const SLOTS_PER_EPOCH: u64 = 32;
+pub const SLOTS_PER_EPOCH: u64 = 32;
 
 /// Contract public methods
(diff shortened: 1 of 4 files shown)


FranklinWaller (approved):

mariocao (changes requested):
I am not entirely sure that using timestamps instead of block heights is the correct approach. Do you mind having a quick discussion before merging this?

My reasoning is that apparently timestamps in NEAR can deviate up to two minutes into the future. See `ACCEPTABLE_TIME_DIFFERENCE` https://github.com/near/nearcore/blob/master/chain/chain/src/chain.rs#L123

mariocao (approved):

PR #67, opened 2023-02-06
Implement token deposits for nodes

jamesondh:
## Motivation

Validators must deposit a minimum amount of tokens and wait for a number of epochs to pass before being eligible to create blocks.

## Explanation of Changes

To facilitate this, the contract was split into three packages:
- mainchain (moved over existing logic)
- token (a new NEP-141 token contract lightly modified from NEAR's examples)
- tests (integration tests using NEAR's workspaces package to test functionality that requires token interactions with the mainchain contract)

Two notable constants need to be determined before mainnet launch:
- `MINIMUM_STAKE` - set to 100 SEDA for now
- `EPOCH_DELAY_FOR_ELECTION` - set to 2, determines how many epochs a validator must wait before being eligible to create blocks (from deposit transaction where the MINIMUM_STAKE is met).

To reduce complexity, nodes are now mapped 1 to 1 with the node operator's NEAR account, so transferring ownership has been removed.

## Testing

Integration tests are included in `contracts/tests` that utilizes [near/workspaces-rs](https://github.com/near/workspaces-rs).

## Related PRs and Issues

N/A


contracts/mainchain/src/consts.rs
```diff
@@ -0,0 +1,7 @@
+use near_sdk::Gas;
+
+pub const BASE_GAS: u64 = 5_000_000_000_000;
+pub const PROMISE_CALL: u64 = 5_000_000_000_000;
+pub const GAS_FOR_FT_ON_TRANSFER: Gas = Gas(BASE_GAS + PROMISE_CALL);
+pub const MINIMUM_STAKE: u128 = 100_000_000_000_000_000_000_000; // 100 SEDA
+pub const EPOCH_DELAY_FOR_ELECTION: u64 = 2;
```
FranklinWaller, line 7: Both MINIMUM_STAKE and EPOCH_DELAY_FOR_ELECTION are not constants, they should always be modifiable by the DAO

contracts/mainchain/src/node_registry.rs
```diff
@@ -0,0 +1,263 @@
...
+    manage_storage_deposit,
+    MainchainContract,
+    MainchainContractExt,
+};
+
+/// Node information
+#[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize, Eq, PartialEq, Debug, Clone, Default)]
+pub struct Node {
+    /// The IP address and port of the node
+    pub socket_address:      String,
```
FranklinWaller, line 26: Would call this a multi_addr to comply with libp2p

contracts/mainchain/src/node_registry.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub epoch_when_eligible: u64,
+}
+
+/// Human-readable node information
+#[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize, Eq, PartialEq, Debug, Clone)]
+pub struct HumanReadableNode {
+    /// The NEAR account id of the node
+    pub account_id:          AccountId,
+    /// The IP address and port of the node
+    pub socket_address:      String,
```
FranklinWaller, line 37: Same here

contracts/mainchain/src/node_registry.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub(crate) fn has_minimum_stake(&self, node: &Node) -> bool {
+        node.balance >= MINIMUM_STAKE
+    }
+
+    pub fn assert_valid_socket_address(&self, socket_address: &String) {
+        for c in socket_address.chars() {
+            assert!(
+                c.is_numeric() || c.is_alphabetic() || c == '.' || c == ':',
+                "Invalid socket address"
+            );
```
FranklinWaller, line 71: I'm not sure if you can validate a multi_addr. The addresses we use in libp2p look like: /ip4/198.51.100.0/tcp/4242/p2p/QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N (https://docs.libp2p.io/concepts/fundamentals/addressing/) but can be longer/shorter

contracts/mainchain/src/node_registry.rs
```diff
@@ -0,0 +1,263 @@
...
+        PromiseOrValue::Value(U128::from(0)) // no refund
+    }
+
+    pub fn internal_withdraw(&mut self, amount: Balance) {
+        assert!(amount > 0, "Withdrawal amount should be positive");
+        let account_id = env::predecessor_account_id();
+        let mut node = self.internal_get_node(&account_id);
+        assert!(node.balance >= amount, "Not enough balance to withdraw");
+
+        // update account
```
FranklinWaller, line 107: Maybe write a todo here, a withdrawal should be requested and only after x amount of epochs it's allowed to fully withdraw

FranklinWaller (changes requested):

contracts/mainchain/src/data_request_test.rs
```diff
@@ -15,13 +15,13 @@ mod tests {
         VMContextBuilder::new()
             .signer_account_id(signer_account_id.parse().unwrap())
             .is_view(false)
-            .attached_deposit(1_400_000_000_000_000_000_000) // required for post_data_request()
+            .attached_deposit(2_140_000_000_000_000_000_000) // required for post_data_request()
```
mariocao, line 18: Where does this storage deposit value comes from?

contracts/mainchain/src/node_registry.rs
mariocao: Then maybe we should mention that it follows the `libp2p` address format.

contracts/mainchain/src/node_registry.rs
mariocao: I wouldn't because there is no way of preventing nodes setting up addresses that don't exist, are offline or belong to other peers.

contracts/mainchain/src/data_request_test.rs
jamesondh: Yes, pretty much trial and error to calculate it: if not enough storage is attached, the transaction will revert and tell you how much you need. The manage_storage_deposit macro handles this

contracts/mainchain/src/dao_test.rs
```diff
@@ -0,0 +1,35 @@
+#[cfg(not(target_arch = "wasm32"))]
+#[cfg(test)]
+mod tests {
```
gluax, line 3: don't need this `mod tests` block here since you have `mod dao_test` in the `lib.rs`. May need to move the `cfg` line to above that statement in the `lib.rs` file though.

contracts/mainchain/src/node_registry_test.rs
```diff
@@ -0,0 +1,143 @@
+#[cfg(not(target_arch = "wasm32"))]
+#[cfg(test)]
+mod tests {
```
gluax, line 3: Same here.

contracts/seda-token/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "seda-token"
+version = "1.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+schemars = { workspace = true }
+near-sdk = { workspace = true, features = ["legacy", "unit-testing"] }
```
gluax, line 11: What's required from `legacy`? We should try to avoid whatever `near` deems as legacy as much as possible IMO, in case they ever deprecate something or etc.

contracts/tests/README.md
```diff
@@ -0,0 +1,13 @@
+SEDA Contracts Integration Tests
+===================
+
+## Building
+To build run in root:
+```make build-wasm
```
gluax, line 6: I think this has since been updated to `make-contracts` right?

gluax (changes requested):

contracts/seda-token/Cargo.toml
jamesondh: not sure, but removing it didn't seem to bork anything 🤷

mariocao (approved):

FranklinWaller (approved):

PR #70, opened 2023-02-09
feat(cli): Add ability to enable/disable mDNS

FranklinWaller:
## Motivation

Some validators want to use mDNS for auto connecting to their own validators. While some don't. This option allows the nodes to enable/disable it through CLI/config.

## Explanation of Changes

Add the option --enable-mdns

## Testing

```seda run --enable-mdns false```

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -51,6 +51,7 @@ pub fn run(seda_server_address: &str, config: NodeConfig, chain_configs: ChainCo
 
         // TODO: Use config for P2P Server
         let mut p2p_server = P2PServer::start_from_config(
+            config.clone(),
             &config.p2p_server_address,
             known_peers,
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: node/src/lib.rs [hunk 1]
```diff
@@ -51,6 +51,7 @@ pub fn run(seda_server_address: &str, config: NodeConfig, chain_configs: ChainCo
 
         // TODO: Use config for P2P Server
         let mut p2p_server = P2PServer::start_from_config(
+            config.clone(),
```
gluax, line 54: I think we should add the P2P config section sooner rather than later. This PR is quite small, so maybe a good place to do it. This way, we can stop tacking on P2P config options to the `NodeConfig` section :).

gluax (approved):

mariocao (approved):

PR #71, opened 2023-02-14
feat(p2p): Add s/Kademlia walks to discovery methods

FranklinWaller:
## Motivation

Implement s/kademlia for peer discovering. This allows us to fill in missing peers in the future (when we build a peer manager)

## Explanation of Changes

Add the kademlia package and add it to the SedaBehaviour. Introduced a new command `seda node peers discover` to trigger a kademlia walk. This command is for now and will probably be removed once the peer manager is in place.

## Testing

Run 3 instances of the node (on different ports with mDNS disabled) Connect node A with B, and B with C. Ask node A to discover new peers using `seda node peers discover`. you'll see that a connection has been established between node A and C by using node B.

### DIFF
--- a/p2p/src/libp2p/behaviour.rs
+++ b/p2p/src/libp2p/behaviour.rs
@@ -16,8 +16,10 @@ use libp2p::{
         ValidationMode,
     },
     identity::Keypair,
+    kad::{store::MemoryStore, Kademlia, KademliaConfig, KademliaEvent},
     mdns::{self},
     swarm::NetworkBehaviour,
+    PeerId,
 };
 
 use super::{super::errors::Result, GOSSIP_TOPIC};
@@ -31,6 +33,8 @@ pub struct SedaBehaviour {
     pub gossipsub: Gossipsub,
     // TODO: change discovery mechanism
     pub mdns:      mdns::async_io::Behaviour,
+
+    pub kademlia: Kademlia<MemoryStore>,
 }
 
 impl SedaBehaviour {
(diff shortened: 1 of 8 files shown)


p2p/src/libp2p/behaviour.rs
```diff
@@ -31,6 +33,8 @@ pub struct SedaBehaviour {
     pub gossipsub: Gossipsub,
     // TODO: change discovery mechanism
     pub mdns:      mdns::async_io::Behaviour,
+
+    pub kademlia_behaviour: Kademlia<MemoryStore>,
```
mariocao, line 37: Why not simply `kademlia`? It is below SedaBehaviour.

```suggestion
    pub kademlia: Kademlia<MemoryStore>,
```

mariocao (approved):

p2p/src/libp2p/mod.rs
```diff
@@ -103,7 +104,10 @@ impl P2PServer {
 
     fn add_peer(&mut self, addr: Multiaddr, peer_id: PeerId) {
         let mut known_peers = self.known_peers.write();
-        known_peers.add_peer(addr, Some(peer_id));
+        known_peers.add_peer(addr.clone(), Some(peer_id));
+
+        self.swarm.behaviour_mut().kademlia.add_address(&peer_id, addr);
```
Thomasvdam, line 109: Probably a Rust thing, but it feels odd that for this new feature you had to add a `clone` to the existing line.

Thomasvdam (approved):

p2p/src/libp2p/mod.rs
FranklinWaller: Yes it's a rust thing 😛

PR #72, opened 2023-02-16
Implement on-chain bn254 verify function

jamesondh:
## Motivation

To verify a signed batch, we need a contract function to verify a bn254 signature. 

## Explanation of Changes

Created `verify` function in `contracts/mainchain/src/verify.rs`. Some code is copied from SEDA's bn254 library including `hash_to_try_and_increment`.

New libraries added to mainchain project:
- zeropool_bn - an BN implementation optimized for NEAR contracts - see https://github.com/zeropoolnetwork/zeropool-near/
- bn254 - using SEDA's bn256 library types only for testing keys and signatures before verifying on-chain
- sha - hashing used in verify function
- near-sys - access to precompile function `alt_bn128_pairing_check`. At some point, rust-sdk-rs [will include this in env](https://github.com/near/near-sdk-rs/issues/644#issuecomment-1131721445) but this PR uses the low-level way of calling it.

## Testing

`test_verify_signed_msg` and `test_verify_aggregate_signatures` are copied from SEDA's bn254 library, replacing the library's verify calls with the contract's.

## Related PRs and Issues

N/A


contracts/mainchain/Cargo.toml
```diff
@@ -9,9 +9,12 @@ rust-version.workspace = true
 crate-type = ["cdylib", "lib"]
 
 [dependencies]
-# Not sure why `unit-testing` feature doesn't work when it's only a dev dep.
+bn254 = "0.0.1"
```
mariocao, line 12: This is a `dev-dependency` right?

jamesondh: correct, moved it 👍

FranklinWaller (approved):

jamesondh:
> I think we might want to expose the try and increment function from the `bn254` function so that we don't need to re-implement it here. Ideally, it would be nice not to have to deal with `Fq` or group operations. What do you think?

Agreed this is better, but `hash_to_try_and_increment()` in our bn254 library returns G1 type from [substrate's bn](https://github.com/paritytech/bn) library, while the reimplementation here returns G1 from [zeropool's bn library](https://github.com/zeropoolnetwork/bn) and implements Borsh serialization. (The precompile function `alt_bn128_pairing_check` uses Borsh for inputs). So we need to either implement Borsh ourselves for this type, or continue to use another bn library. Both zeropool and aurora maintain a `bn` library with Borsh, but [Aurora's may be better maintained](https://github.com/near/nearcore/pull/3971#issuecomment-817676944). We may want to consider replacing substrate's bn in our `bn254` library as well with one of these as to minimize dependencies.

contracts/mainchain/src/verify.rs
```diff
@@ -0,0 +1,33 @@
+use bn254::hash_to_try_and_increment;
+use near_sdk::{borsh::BorshSerialize, near_bindgen};
+use near_sys::alt_bn128_pairing_check;
+use zeropool_bn::{Group, G1, G2};
```
mariocao, line 4: I think that after the last PR from `bn254` to use `zeropool` instead of `parity` we might not need to import `G1` and `G2`. We should be able to use the `from_compressed` and `to_compressed` directly from the `bn254`.

jamesondh: We have to import G2 for `n_g2` anyways, so maybe it's better to create a helper function on `bn254` to format the pairing values here

jamesondh: I see this options:
- keep the zeropool bn import here since we need G2:one()
- write a small helper function on bn254 to return G2:one()
- write a helper function on bn254 to format the pairing values, which would require adding Borsh as a dependency

gluax: All things should be able to be done from our bn254 crate.

I lean towards option 3, myself. We can implement borsh serialize and deserialize on our own type manually that just calls to the one from zeropool.

jamesondh: https://github.com/sedaprotocol/bn254/pull/4

mariocao (approved):
Now it looks really clean!!! 💯 

But we have the error that @mennatabuelnaga was mentioned on her PR https://github.com/sedaprotocol/seda-rust/pull/75

PR #75, opened 2023-02-20
feat: create a random key on node startup if not found in env

mennatabuelnaga:
- Renames the preexisting `SEDA_SECRECT_KEY` to `CHAIN_SECRET_KEY`.
- Checks for a `SEDA_SECRET_KEY` in .env, if not found it generates a random secret key using `rand` and `bn254` crates, then writes it to a file `random_sk`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,6 @@
 SEDA_LOG_FILE_PATH = "some path of where to store seda log files"
 SEDA_NEAR_RPC_URL = "https://rpc.testnet.near.org"
+CHAIN_SECRET_KEY = "some secret key"
 SEDA_SECRET_KEY = "some secret key"
 SEDA_SERVER_ADDRESS = "127.0.0.1"
 SEDA_SERVER_PORT = 12345
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 SEDA_LOG_FILE_PATH = "some path of where to store seda log files"
 SEDA_NEAR_RPC_URL = "https://rpc.testnet.near.org"
-CHAIN_SECRET_KEY = "some secret key"
+SEDA_CHAIN_SECRET_KEY = "some secret key"
 SEDA_SECRET_KEY = "some secret key"
 SEDA_SERVER_ADDRESS = "127.0.0.1"
 SEDA_SERVER_PORT = 12345
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ async-trait = "0.1"
 base64 = "0.13"
 bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
 borsh = { version = "0.9", default-features = false }
+bn254 = "0.0.1"
 clap = { version = "4.1", default-features = false }
 clap_complete = { version = "4.1", default-features = false }
 clap-markdown = { version = "0.1", default-features = false }
@@ -89,4 +90,3 @@ uint = { version = "0.8", default-features = false }
 wasmer = { version = "2.3", default-features = false }
 wasmer-wasi = { version = "2.3", default-features = false }
 workspaces = "0.7"
-zeropool-bn = "0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ near-primitives = "0.15"
 near-sdk = { version = "4.0", default-features = false }
 near-units = "0.2"
 parking_lot = "0.12"
+rand = "0.8.5"
 reqwest = "0.11"
 rusqlite = { version = "0.28", features = ["bundled"] }
 schemars = "0.8"
@@ -88,3 +89,4 @@ uint = { version = "0.8", default-features = false }
 wasmer = { version = "2.3", default-features = false }
 wasmer-wasi = { version = "2.3", default-features = false }
 workspaces = "0.7"
+zeropool-bn = "0.5"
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -86,3 +87,4 @@ uint = { version = "0.8", default-features = false }
 wasmer = { version = "2.3", default-features = false }
 wasmer-wasi = { version = "2.3", default-features = false }
 workspaces = "0.7"
+zeropool-bn = "0.5"
```
gluax, line 90: Are we using `zeropool-bn` in place of our `bn254` repo for now?

node/src/lib.rs
```diff
@@ -33,6 +36,14 @@ pub fn run(seda_server_address: &str, config: NodeConfig, p2p_config: P2PConfig,
     let system = System::new();
     // Initialize actors inside system context
     system.block_on(async {
+        // Checks if there is a SEDA_SECRET_KEY env variable
```
gluax, line 39: Probably a future thing: but this should maybe a function, and then we can have users do this from the CLI at some point by calling that function. It doesn't need to be done now; just food for thought.

node/src/lib.rs
```diff
@@ -33,6 +36,14 @@ pub fn run(seda_server_address: &str, config: NodeConfig, p2p_config: P2PConfig,
...
+        // If not, it generates a new random one and saves it into a file
+        if std::env::var("SEDA_SECRET_KEY").is_err() {
+            let rng = &mut rand::thread_rng();
+            let sk = Fr::random(rng);
+            let mut f = File::create("./random_sk").expect("Unable to create file");
```
FranklinWaller, line 44: Could you make this a constant and name it seda_key instead of random_sk

Cargo.toml
mariocao: I think we shouldn't.

Cargo.toml
```diff
@@ -44,8 +44,8 @@ opt-level = "z"
 actix = "0.13"
 async-trait = "0.1"
 base64 = "0.13"
-bn254 = "0.0"
 borsh = { version = "0.9", default-features = false }
+bn254 = "0.0.1"
```
gluax, line 48: We should do `bn254 = { git = "https://github.com/sedaprotocol/bn254/", branch = "main" }` for now.

This way, we don't have to publish new versions while we update :).
Better yet, we use the git upstream branch for dev versions, and for release, we use the released version. I'm not sure cargo lets us do that. But worth looking into.

node/src/lib.rs
```diff
@@ -28,11 +31,19 @@ use crate::app::Shutdown;
 pub mod test {
     mod event_queue_test;
 }
-
+const SK_PATH: &str = "./seda_key";
```
gluax, line 34: We should always do it locally to where the application runs from like this. It could be a config field and env var. Or it should be like our config file that writes to the standard OS config directory. Or an argument could also be made for the `HOME` equivalent OS application directory. 

@mariocao @FranklinWaller, any thoughts here?

FranklinWaller: A config field is always nice, but I do prefer the default to be ./seda_key instead of scattered around

gluax: Let's make it a config field that defaults to that for now then!

Cargo.toml
mariocao: Great!! :)

.env.example
```diff
@@ -1,5 +1,6 @@
 SEDA_LOG_FILE_PATH = "some path of where to store seda log files"
 SEDA_NEAR_RPC_URL = "https://rpc.testnet.near.org"
+CHAIN_SECRET_KEY = "some secret key"
```
mariocao, line 3: We should add the `SEDA` prefix, right?
To be homogenous with the rest of variables.

README.md
```diff
@@ -100,7 +100,7 @@ SEDA configuration uses the following ENV variables if they exist.
 | `SEDA_CONFIG_PATH`    | Defines an alternative path for the seda configuration file to be.                                                                 |
 | `SEDA_LOG_FILE_PATH`  | Overwrites the config `logging.log_file_path` field.                                                                               |
 | `SEDA_NEAR_RPC_URL`   | Overwrites the config `near_chain.chain_rpc_url` field.                                                                            |
-| `SEDA_SECRET_KEY`     | Overwrites the config `node.secret_key` field.                                                                                     |
+| `CHAIN_SECRET_KEY`     | Overwrites the config `node.secret_key` field.                                                                                     |
```
mariocao, line 103: `SEDA_SECRET_KEY` is missing right? :)

node/src/lib.rs
mariocao: I think it makes sense. For example, calling a CLI command for generating a secret key and showing the derived `ed25519` public key. This could be used for associating a stake to a validator identity.

Cargo.toml
mennatabuelnaga: For some weird reason, setting it to the main branch throws an error when building wasm binaries: `error: linking with `rust-lld` failed: exit status: 1` ` rust-lld: error: function signature mismatch: promise_then`

gluax (approved):

gluax:
Looks good to me, but I noticed a file named `seda_key` was pushed. We should probably `gitignore` those :)

mariocao:
> Looks good to me, but I noticed a file named `seda_key` was pushed. We should probably `gitignore` those :)

That should be part of the PR

node/src/lib.rs
FranklinWaller: Can we do this now tbh? Because it shouldn't be too much work and it cleans it up :)

mariocao (approved):

gluax (approved):

PR #76, opened 2023-02-21
Store bn254 public key with validator on contract

jamesondh:
## Motivation

When a new batch is posted, we must verify a aggregate signature. This PR requires node operators to store their bn254 public key on the contract when registering so that they may be fetched when doing batch signature verification.

## Explanation of Changes

`register_node()` now takes two additional arguments: a bn254 public key, and a signature using that key of the caller's NEAR account name (to prove that the bn254 key owner wants to use a specific NEAR account). That signature is then verified and stored with the `Node`.

## Testing

`register_and_get_node()` is updated to include the two new arguments for registering a node. 

## Related PRs and Issues

Blocked by https://github.com/sedaprotocol/seda-rust/pull/72.

Future or related issues may be verifying signatures are unique for every NEAR address, or allowing changing keys.


contracts/mainchain/src/verify.rs
```diff
@@ -0,0 +1,136 @@
...
+            .and_then(|_| arbitrary_string_to_g1(&s).ok())
+    });
+
+    // Return an error if no valid point was found
+    Ok(point.unwrap())
+}
+
+#[near_bindgen]
+impl MainchainContract {
+    pub fn verify(&mut self, message: Vec<u8>, signature: Vec<u8>, public_key: Vec<u8>) -> bool {
```
mariocao, line 113: I would rename the function to `bn254_verify` to avoid confusions.

contracts/mainchain/src/node_registry.rs
```diff
@@ -37,6 +38,7 @@ pub struct HumanReadableNode {
     pub multi_addr:          String,
     pub balance:             Balance,
     pub epoch_when_eligible: U64,
+    pub public_key:          Vec<u8>,
```
mariocao, line 41: Same comment here, I would document that this public key is from `bn254` or even rename the field as `bn254_public_key`.  Wdyt?

PR #77, opened 2023-02-22
Feat/promise rejections

gluax:
## Motivation

So that errors within the runtime turn into promise rejections rather than halting the WASM execution.

## Explanation of Changes

Moved the `Bytes` type and its methods to the SDK(which also allowed some code cleanup in WASM files).
Then implemented, the `From` method for `Result<T: ToBytes, E: Error>` as well as  `Result<Option<T: ToBytes>, E: Error>`.

Most of the promise action handlers have now become one-liners.

The one exception is that `CallSelf` has not been updated to leverage this system.

That is because it has many more details and intricacies. Some of those errors mean we could not reject a promise because it means a runtime would fail to be started. While others would mean that we fail to capture `stdout`, `stderr`, etc. This code section would be better addressed in the PR that fixes that problem.

Finally, there's been a bunch of `TODO`s added in the SDK and other places. Much of this code can be cleaned up to make working with it easier(especially from the WASM side of things). However, in an effort to keep smaller PRs, these can be done in a Runtime clean-up PR.

## Testing

Added a new test with an intentionally failing HTTP request to make sure that error turns into a promise rejection.

## Related PRs and Issues

N/A

### DIFF
--- a/contracts/tests/src/lib.rs
+++ b/contracts/tests/src/lib.rs
@@ -1,3 +1,6 @@
+#[cfg(target_family = "unix")]
 mod seda_token;
+#[cfg(target_family = "unix")]
 mod staking;
+#[cfg(target_family = "unix")]
 mod utils;
--- a/runtime/core/src/runtime.rs
+++ b/runtime/core/src/runtime.rs
@@ -176,7 +176,8 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
 
                         let execution_result = vm_context.result.lock();
                         next_promise_queue = next_queue.lock().clone();
-                        promise_queue_mut.queue[index].status = PromiseStatus::Fulfilled(execution_result.clone());
+                        promise_queue_mut.queue[index].status =
+                            PromiseStatus::Fulfilled(Some(execution_result.clone()));
                     }
 
                     // Just an example, delete this later
@@ -189,20 +190,7 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
                     }
 
                     PromiseAction::DatabaseGet(db_action) => {
-                        // TODO discuss how we should handle options for promise rejections!
-                        let result = self
-                            .host_adapter
-                            .db_get(&db_action.key)
-                            .await
-                            .map_err(|e| RuntimeError::NodeError(e.to_string()))?;
-
-                        match result {
-                            Some(r) => {
-                                promise_queue_mut.queue[index].status =
-                                    PromiseStatus::Fulfilled(r.to_string().into_bytes())
-                            }
-                            None => promise_queue_mut.queue[index].status = PromiseStatus::Rejected(vec![]),
-                        }
+                        promise_queue_mut.queue[index].status = self.host_adapter.db_get(&db_action.key).await.into();
                     }
 
                     PromiseAction::Http(http_action) => {
@@ -243,7 +231,7 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
                             .into();
                     }
                     PromiseAction::P2PBroadcast(p2p_broadcast_action) => {
-                        // TODO do we return no promise rejection or success here intentionally?
+                        // TODO we need to figure this out at some point.
                         p2p_command_sender_channel
                             .send(P2PCommand::Broadcast(p2p_broadcast_action.data.clone()))
                             .await?;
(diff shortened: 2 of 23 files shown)


gluax:
Draft PR since I realized I forgot to handle something in the promise action for `CallSelf` :)

jamesondh (approved):

contracts/tests/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(target_family = "unix")]
 mod seda_token;
+#[cfg(target_family = "unix")]
 mod staking;
+#[cfg(target_family = "unix")]
```
FranklinWaller, line 5: why do we need this exactly?

runtime/core/src/runtime.rs
```diff
@@ -27,6 +27,13 @@ pub struct VmResult {
     pub exit_code: u8,
 }
 
+// TODO: can I move this trait to the sdk?
```
FranklinWaller, line 30: I wouldn't bother with it tbh, it's not going to be used anywhere else but this place

runtime/core/src/runtime.rs
```diff
@@ -231,19 +212,17 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
...
+                            .host_adapter
                             .trigger_event(trigger_event_action.event.clone())
                             .await
-                            .map_err(|e| RuntimeError::NodeError(e.to_string()))?;
-
-                        promise_queue_mut.queue[index].status = PromiseStatus::Fulfilled(vec![]);
+                            .into();
                     }
                     PromiseAction::P2PBroadcast(p2p_broadcast_action) => {
+                        // TODO we need to figure this out at some point.
```
FranklinWaller, line 225: Maybe clarify a bit more 😛 For handling success/error situations using channels

runtime/sdk/src/promises/promise.rs
```diff
@@ -11,12 +13,32 @@ pub enum PromiseStatus {
     Pending,
 
     /// The promise completed
-    Fulfilled(Vec<u8>),
+    Fulfilled(Option<Vec<u8>>),
 
     /// There was an error executing this promise
+    // TODO: Is there ever a case where Rejected isn't a string?
```
FranklinWaller, line 19: for http it could also be a object that gives the reason why it failed (but that's encoded in a string)

wasm/test/promise-wasm-bin/src/main.rs
```diff
@@ -165,16 +156,18 @@ fn bn254_sign_test() {
...
 
+// TODO: Something to include in our SDK? Or bn254 lib. Or use hex crate.
 fn decode_hex(s: &str) -> Result<Vec<u8>, ParseIntError> {
     (0..s.len())
         .step_by(2)
         .map(|i| u8::from_str_radix(&s[i..i + 2], 16))
         .collect()
 }
 
+// TODO: Something to include in our SDK? Or bn254 lib. Or use hex crate.
```
FranklinWaller, line 170: SDK would be nice, high chance we use it later

contracts/tests/src/lib.rs
gluax: The `contracts/tests` module depends on `workspaces`, which only works on UNIX systems. So when developing in our repo on windows, `rust-analyzer` fails to run for the project, and running the contract tests also results in a compilation error.

runtime/core/src/runtime.rs
gluax: yea there's probably a better way to clean it up than moving that trait. I'll adjust the todo.

runtime/sdk/src/promises/promise.rs
gluax: Hmmm, ok, I will leave that todo for now, but make a note of that. Many of these todos are me thinking about how we might be able to clean some stuff up eventually.

wasm/test/promise-wasm-bin/src/main.rs
gluax: I'm gonna leave this for another PR where we clean up the runtime a bit though :)

mariocao: I was tempted to use `hex` crate, but because it was used once and it was a test and left it like that.

mariocao (approved):

PR #78, opened 2023-02-27
feat(p2p): Add peer manager

FranklinWaller:
## Motivation

We want the node to always try to connect to a certain amount of peers using different discovery methods. This implements a manager that allows for this functionality. 

## Explanation of Changes

This implements a "waterfall" peer manager that tries to connect to peers in the following order:
* Manual peers
* mDNS peers
* Chain peers (not implemented)
* Kademlia peers
 
It goes one by one and checks if there are any peers to connect to. If not it will go to the next discovery method. If all methods are exhausted it will ask kademlia to search the network for new peers.
If a peer is not reachable it will be put inside the cooldown list, which the node will not retry for a certain amount of time. After which the node retries the connection.

## Testing

The discovery methods are untouched, you can test this by using mDNS (spin up two nodes and wait for them to find each other) or use manual peers. If you want to test kademlia you have to disable all methods but kademlia and connect the nodes together through the CLI.

### DIFF
--- a/p2p/src/libp2p/discovery_status.rs
+++ b/p2p/src/libp2p/discovery_status.rs
@@ -1,0 +1,219 @@
+use std::{collections::HashMap, ops::Add, time::SystemTime};
+
+use libp2p::{Multiaddr, PeerId};
+use seda_config::P2PConfig;
+
+use super::peer_list::{ConnectionType, PeerInfo};
+use crate::PeerList;
+
+/// The manager should use this
+/// The manager should not check the config but just have a switch statement
+/// This struct will check all connections and config which discovery method
+/// should be used
+
+pub struct DiscoveryStatus {
+    p2p_config: P2PConfig,
+
+    pub connected_peers: PeerList,
+
+    /// Peers who should not be connected to until the cooldown period has ended
+    /// This prevents the manager to try to connect to the same failing peer
+    cooldown_addrs: HashMap<Multiaddr, SystemTime>,
+
+    // Peers that have been found by a discovery mechanism
+    // but don't have to be connected, we can use this if we want to add them
+    found_manual_peers:   PeerList,
+    found_chain_peers:    PeerList,
+    found_mdns_peers:     PeerList,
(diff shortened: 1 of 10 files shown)


p2p/src/libp2p/peer_list.rs
```diff
@@ -68,4 +130,8 @@ impl PeerList {
     pub fn get_all(&self) -> HashMap<PeerId, Multiaddr> {
         self.peer_to_addr.clone()
     }
+
+    pub fn get_all_info(&self) -> HashMap<Multiaddr, PeerInfo> {
+        self.addr_to_peer.clone()
```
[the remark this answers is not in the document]
gluax: We can do this in a follow up PR when we have the bandwith.

p2p/src/libp2p/discovery_status.rs
```diff
@@ -0,0 +1,212 @@
...
+            return false;
+        }
+
+        true
+    }
+
+    /// Gets the discovery method the manager should use
+    /// Based on priority and maximum allowed peers from that specific discovery
+    /// method
+    /// * `skip` - Allows to skip any higher priorty then the one given (if for
```
mariocao, line 82:
```suggestion
    /// * `skip` - Allows to skip any higher or equal priority then the one given (if for
```

p2p/src/libp2p/mod.rs
```diff
@@ -53,126 +57,150 @@ impl P2PServer {
...
+        match current_discovery_method {
+            ConnectionType::Manual => self.search_manual_peers(),
+            ConnectionType::MDns => self.search_mdns_peers(),
+            ConnectionType::Chain => {
+                // TODO: Add chain peers
+                self.search_new_peer(Some(ConnectionType::Chain));
+            }
+            ConnectionType::Kademlia => self.search_kademlia_peers(),
+            ConnectionType::None => {
+                tracing::debug!("No new peers found");
```
mariocao, line 97: It is not only that new peers cannot be found but that maybe no new peers are needed, right?

PR #82, opened 2023-03-01
feat(contracts): implement basic `post_signed_batch` method

jamesondh:
## Motivation

After getting the the merkle root of accumulated data requests on the mainchain contract via the view function `compute_merkle_root()`, nodes that are part of the current committee sign it with their bn254 key, and then a leader submits the aggregated signatures to the contract to be verified and then push a new batch. This PR implements the function `post_signed_batch()` with basic functionality.

## Explanation of Changes

- Implemented `post_signed_batch()`, which takes an aggregated signature, checks if all signers are in the current epoch's committee, verifies the signature, pushes the batch to the contract, and clears the data request accumulator
- Renamed block to batch on the contract where applicable

Assumptions/caveats:
- any node operator eligible for the current epoch may be included in the aggregate signature
- node operators may abstain from providing a signature
- batches are not yet linked to slots (e.g. no 1 batch per 1 slot requirement)
- if data requests are added after the signature aggregation, the function will fail (since data requests are not yet associated with the slot in which they are eligible to be included in a batch)

## Testing

An integration test `test_post_signed_batch()` tests a scenario in which two nodes are registered, a few data requests are pushed to the contract, and then the signed batch is posted.


contracts/mainchain/src/batch.rs
```diff
@@ -0,0 +1,107 @@
...
+            "Data request accumulator is empty"
+        );
+
+        // check if all signers are eligible for this epoch and match the aggregate
+        // signature initialize with the first signer's public key
+        self.assert_eligible_for_current_epoch(&signers[0]);
+        let mut aggregate_public_key_check =
+            PublicKey::from_compressed(self.nodes.get(&signers[0]).unwrap().bn254_public_key).unwrap();
+        // add the rest of the signer's public keys
+        for signer in signers.iter().skip(1) {
```
gluax, line 61: Can we write a comment here explaining why we skip 1?

contracts/tests/Cargo.toml
```diff
@@ -18,6 +18,7 @@ seda-mainchain = { path = "../mainchain" }
 serde = { workspace = true }
 serde_json = { workspace = true }
 tokio = { workspace = true }
+rand = "0.8.5"
```
gluax, line 21: From the workspace?

gluax (changes requested):

gluax (approved):

mennatabuelnaga (approved):

mariocao (approved):

PR #83, opened 2023-03-01
Feat/wasm shared memory

gluax:
## Motivation

To add a memory storage that lasts for the entirety of the lifetime of the node.

## Explanation of Changes

Mostly copy-paste of existing memory code. However, this is an `RwLock` (we should consider updating the other one to a `RwLock` as well?), and it is created with the creation of the `App` and shared from there.

## Testing

Originally tried to write a test at the `Node` level, but that is much harder to do now. Instead, I added a test that memory is shared properly over different `Runtime`s to prove that it's shared instead.

We can eventually add `Node` tests.

## Related PRs and Issues

N\A


jamesondh (approved):

FranklinWaller (approved):

FranklinWaller:
The non shared memory doesn't have to be RwLock, unless we are doing some parallel stuff in one runtime instance.

mariocao (approved):

PR #85, opened 2023-03-03
feat(node): sign batch

gluax:
## Motivation

To sign batches and send them to the rest of the network.

## Explanation of Changes

Moved the `KeyPair` to actually be on the `NodeConfig`, and auto-derived the public key.
This is for the convenience of accessing the key pair from the config instead of having to clone it and build it when needed.

Adds the batch signing to the consensus wasm module.
**Notes**:

- I had to store the private_key in memory, for now, to get around the env bug.
- I also had to ignore the  `PromiseResult` of the broadcast message in the batch signer logic. This is due to us not having implemented that logic yet.

Adds a batch tick(currently set to 5 seconds, this will need to be changed in the future).

## Testing

Run the test RPC server.
Run two nodes that connect to each other, and watch them send messages to each other :).

Here's an example of it running on two nodes locally :).

image

## Related PRs and Issues

~~Blocked by #90.~~

~~Would like to see this [pr](https://github.com/sedaprotocol/bn254/pull/6) merged before this one, so I can set the branch on bn254 back to `main`.~~


node/src/host/batch_tick.rs
```diff
@@ -0,0 +1,30 @@
...
+    const BATCH_TICK_INTERVAL: u64 = 1000;
+}
+
+impl<HA: HostAdapter> Handler<BatchTickManager> for Host<HA> {
+    type Result = ();
+
+    fn handle(&mut self, msg: BatchTickManager, ctx: &mut Self::Context) -> Self::Result {
+        let event = Event::new("BatchChainTick", EventData::BatchChainTick);
+        if let Some(app) = self.app_actor_addr.as_ref() {
+            app.do_send(AddEventToQueue::new(event));
```
mariocao, line 25: Why not `AddEventToQueue{event}`?

node/src/event_queue_handler.rs
```diff
@@ -11,6 +11,12 @@ pub struct AddEventToQueue {
     pub event: Event,
 }
 
+impl AddEventToQueue {
+    pub fn new(event: Event) -> Self {
+        AddEventToQueue { event }
+    }
+}
+
```
mariocao, line 19: We could use directly `AddEventToQueue{event: event_obj}`, right?

runtime/core/src/runtime.rs
```diff
@@ -244,6 +245,11 @@ impl<HA: HostAdapter> RunnableRuntime for Runtime<HA> {
                             .await
                             .into();
                     }
+                    PromiseAction::Rpc(RpcAction { url, method, args }) => {
```
FranklinWaller, line 248: Do we need RPC? Can't we use Http?

gluax: We could. However, I find the QOL of keeping them separate better. The QOL is for both the WASM side and the Rust side. The Rust side similarly lets us use a library made for interacting with RPCs. So it's a win-win as far as I see it.

FranklinWaller: I think having 2 different Http methods to maintain is a bit weird but It's ok. I would at least remove any reliance on WebSockets. RPCs from chains are not guaranteed to have web socket support, limiting the amount of RPCs a user can pick is not desired.

gluax: You got it. My bad, all the ones I've encountered were WS only. So I'll remove that check.

gluax:
The PR changes line numbers are smaller than Github reports. For some reason, it's counting the mock RPC changes already merged.

Cargo.toml
```diff
@@ -44,7 +45,7 @@ opt-level = "z"
 actix = "0.13"
 async-trait = "0.1"
 base64 = "0.13"
-bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
+bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "feat/derives" }
```
mariocao, line 48: I approved the PR on the `bn254` repo. :)

crypto/src/crypto_test.rs
```diff
@@ -1,39 +1,34 @@
...
 }
 
-#[tokio::test]
-async fn generate_bn254_pair() {
-    generate_test_sk().await;
-    let bn_pair: Bn254KeyPair = derive_bn254_key_pair(TEST_SK_PATH, 1).expect("Couldn't derive bn254 key pair");
+#[test]
+fn generate_bn254_pair() {
+    generate_test_sk();
+    let bn_pair = KeyPair::derive(TEST_SK_PATH, 1).expect("Couldn't derive bn254 key pair");
```
mariocao, line 20: Should this `KeyPair` be the one from the crypto module? (and not from ed25519_dalek)
```suggestion
    let bn_pair = KeyPair::derive(TEST_SK_PATH, 1).expect("Couldn't derive bn254 key pair");
```

gluax: It is the one from ours! Notice `KeyPair`(ours) vs. `Keypair` (`ed25519_dalek`).

mariocao (approved):

FranklinWaller (approved):

runtime/sdk/src/wasm/chain_interactor.rs
```diff
@@ -10,7 +10,13 @@ pub fn chain_view<S: ToString>(chain: Chain, contract_id: S, method_name: S, arg
     }))
 }
 
-pub fn chain_call<S: ToString>(chain: Chain, contract_id: S, method_name: S, args: Vec<u8>, deposit: u128) -> Promise {
+pub fn chain_call<C: ToString, M: ToString>(
+    chain: Chain,
+    contract_id: C,
+    method_name: M,
```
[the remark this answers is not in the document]
gluax: It's purely convenient for user experience. It has two separate generics that implement `ToString`. Reason being if I used one generic it has to be the same type in both places for that generic. Having two allows you to pass two separate types that implement the trait.

PR #88, opened 2023-03-06
feat(delegate-cli): Add top-up commando for topping up NEAR to a spec…

FranklinWaller:
…ific account id

## Motivation

Add a new separate CLI interface which is just for staking and topping up. This PR only includes topping up and the general structure to make the CLI work standalone.

## Explanation of Changes

Add a new config to the config package and a new CLI project called "delegate-cli" (Binary is called "seda-delegate"), currently only top-up is supported

## Testing

Create a .env file with the following content:

`NEAR_ENV=testnet near login`

you can find your credentials in `.near-credentials`

```
ACCOUNT_SECRET_KEY = "ed25519:YOUR_KEY_HERE"
SIGNER_ACCOUNT_ID = "YOUR_ACCOUNT_ID"
```
then you can run

`make run-build-delegate top-up franklinwaller.testnet 0.001`

This will transfer 0.001N to the account

### DIFF
--- a/config/src/configs/delegate.rs
+++ b/config/src/configs/delegate.rs
@@ -1,0 +1,112 @@
+use std::sync::Arc;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{env_overwrite, ConfigError};
+#[cfg(feature = "delegate-cli")]
+use crate::{merge_config_cli, Config, Result};
+
+#[cfg(feature = "delegate-cli")]
+#[derive(clap::Args)]
+/// The configuration for the seda engine.
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct PartialDelegateConfig {
+    /// An option to override the validator secret key config value.
+    #[arg(long)]
+    pub validator_secret_key: Option<String>,
+    /// An option to override the account secret key config value. (Used to sign
+    /// top-ups and stakings)
+    #[arg(long)]
+    pub account_secret_key:   Option<String>,
+    /// An option to override the signer account ID config value.
+    #[arg(long)]
+    pub signer_account_id:    Option<String>,
+    /// An option to override the delegate contract id
+    #[arg(long)]
+    pub delegate_contract_id: Option<String>,
+    /// An option to override the RPC URL
+    #[arg(long)]
+    pub rpc_url:              Option<String>,
+}
+
+#[cfg(feature = "delegate-cli")]
+impl PartialDelegateConfig {
+    pub fn to_config(self, cli_options: Self) -> Result<DelegateConfig> {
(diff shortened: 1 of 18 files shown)


config/src/configs/delegate.rs
```diff
@@ -0,0 +1,97 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{env_overwrite, ConfigError};
+#[cfg(feature = "delegate-cli")]
+use crate::{merge_config_cli, Config, Result};
+
+#[cfg(feature = "delegate-cli")]
+#[derive(clap::Args)]
+/// The configuration for the seda engine.
+#[derive(Debug, Default, Serialize, Deserialize)]
```
gluax, line 12: Instead of making a whole new Config struct, this could be another Partial config struct that turns into a `NodeConfig`.
I think we already have an example of this kind of thing in the `NodeConfig` if not we did have it in an older commit.
I believe this would be cleaner than having the same information in multiple config blocks in our files, or multiple config files with similar information.

delegate-cli/src/main.rs
```diff
@@ -0,0 +1,19 @@
+use clap::Parser;
+use seda_config::{Config, PartialDelegateConfig};
+
+use crate::cli::CliOptions;
+
+mod cli;
+
+fn main() {
```
gluax, line 8: Why do a separate binary? Why not have one binary as a single source of truth?

We can have a `Delegate` subcommand, i.e., `seda delegate top-up...`.

FranklinWaller: It's to not have the secret key that holds the SEDA tokens in the same env

gluax: We should add checks that deny running this tool if the `env` contains variables we don't want, likewise for the regular `seda` binary. It's a more concrete way of stopping people from running this in a way that makes them vulnerable. Plus, it's a failsafe since it relies on code, not people understanding the software usage.

delegate-cli/src/cli/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+pub struct CliOptions {
+    #[command(flatten)]
+    pub delegate_config: PartialDelegateConfig,
+    #[command(subcommand)]
+    pub command:         Command,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum Command {
+    TopUp(commands::top_up::TopUp),
```
mariocao, line 23: Add some comment to have the CLI `--help` show some command info.

```suggestion
    /// Top-up a target account by a given amount in NEAR
    TopUp(commands::top_up::TopUp),
```

mariocao (approved):

gluax (approved):

PR #89, opened 2023-03-08
feat(contracts): implement committee bootstrapping

jamesondh:
## Motivation

Because committees are selected a few epochs in advance to allow nodes to prepare, we need a bootstrapping mechanism to select the first few committees once there are enough active nodes.

## Explanation of Changes

- Explicitly track `active_nodes` and `inactive_nodes` to make it easier to select committee members
- Created `process_epoch()` function, which does 2 things:
  1. Moves pending nodes from `inactive_nodes` to `active_nodes` once their activation epoch is reached
  2. Selects committee members for (current epoch + `EPOCH_COMMITTEES_LOOKAHEAD`). The contract is initialized with `bootstrapping_phase = true`, so once there are `VALIDATOR_COMMITTEE_SIZE` active nodes, it selects committees for multiple epochs and then sets bootstrapping_phase to false.

## Testing

Tests have been updated to call `process_epoch()` where `post_signed_batch()` is called. The `post_signed_batch()` test in `batch_test.rs` also checks to see if the bootstrapping creates 3 committees and that each committee has `SLOTS_PER_EPOCH` members.

## Related PRs and Issues

Still need to implement:
- Only slot leader can call `post_signed_batch()`
- Restrictions on withdrawing while an active node or pending node
- Committee selection algorithm instead of round-robin

Optimizations are also welcome if you spot one :)

### DIFF
--- a/contracts/src/batch.rs
+++ b/contracts/src/batch.rs
@@ -56,14 +56,20 @@ impl MainchainContract {
         // eligible for this batch while also verifying individual eligibility
 
         // 1. initialize with the first signer
-        self.assert_eligible_for_current_epoch(&signers[0]);
+        assert!(
+            self.active_nodes.get(&signers[0]).is_some(),
+            "Node is not active"
+        );
         let mut aggregate_public_key_check =
-            PublicKey::from_compressed(self.nodes.get(&signers[0]).unwrap().bn254_public_key).unwrap();
+            PublicKey::from_compressed(self.active_nodes.get(&signers[0]).unwrap().bn254_public_key).unwrap();
         // 2. add the rest of the signers' public keys
         for signer in signers.iter().skip(1) {
-            self.assert_eligible_for_current_epoch(signer); // TODO: store in a vector of eligible signers for this epoch
+            assert!(
+                self.active_nodes.get(&signer).is_some(),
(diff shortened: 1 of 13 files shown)


gluax (approved):

contracts/src/batch.rs
```diff
@@ -56,14 +64,14 @@ impl MainchainContract {
...
+            PublicKey::from_compressed(self.active_nodes.get(&signers[0]).unwrap().bn254_public_key).unwrap();
         // 2. add the rest of the signers' public keys
         for signer in signers.iter().skip(1) {
-            self.assert_eligible_for_current_epoch(signer); // TODO: store in a vector of eligible signers for this epoch
+            assert!(self.committees[0].contains(signer), "Node is not active");
             let signer_public_key =
-                PublicKey::from_compressed(self.nodes.get(signer).unwrap().bn254_public_key).unwrap();
+                PublicKey::from_compressed(self.active_nodes.get(signer).unwrap().bn254_public_key).unwrap();
             aggregate_public_key_check = aggregate_public_key_check + signer_public_key;
         }
```
mariocao, line 76: We could have done this with a `fold` function, right?

mariocao (approved):

PR #90, opened 2023-03-09
Feat: RPC Mock

gluax:
## Motivation

To add an RPC testing mock stuff.

## Explanation of Changes

Adds a new binary, `seda-debugger`, which can spin up and stop the test RPC server.
Adds a bunch of new `Makefile` commands around the debugger.
It also spins up the test RPC server and stops it automatically when running `make test`.

**NOTE**:

All data for the test RPC is currently statically mocked. However, in the future, this should be dealt with as necessary.

## Testing

Tested manually that our calls all worked with the Test RPC server.
i.e. in the config file, I did the following

```toml
[chains.near]
chain_rpc_url = 'http://127.0.0.1:4657'
```

before running our node commands.
## Related PRs and Issues

N/A


jamesondh (approved):

mariocao (approved):
💯

PR #91, opened 2023-03-12
feat(contracts): deposit on a ed25519 key

jamesondh:
## Motivation

To ensure the SEDA node does not have access to tokens, we must allow depositing/withdrawing tokens to a specific ed25519 key.

## Explanation of Changes

- New global variables:
  - `nodes_by_ed25519_public_key` mapping to fetch the AccountId of a given ed25519 key (used in deposit and withdraw functions)
  - `depositors` mapping to store all nodes a depositor has funds with
- Deposit and withdraw functions now accept a ed25519 key argument instead of assuming deposit/withdraw into one's own node
- Refactored `test_utils.rs` to include predefined test accounts with bn254 and ed25519 keys and improve test readability

## Testing

New tests in `node_registry_tests.rs`:
- `withdraw_wrong_account()` to try withdrawing from a node where the caller has no funds deposited
- `deposit_withdraw_one_node_two_depositors()` to test multiple accounts depositing/withdrawing tokens on a node
- `deposit_withdraw_two_nodes_one_depositor()` to test a depositor adding and removing tokens on multiple nodes

## Related PRs and Issues

N/A

### DIFF
--- a/contracts/src/node_registry.rs
+++ b/contracts/src/node_registry.rs
@@ -104,8 +104,7 @@ impl MainchainContract {
             let depositor_account_id = env::signer_account_id();
 
             // subtract from user balance and add to contract balance
-            let new_user_balance = self.token.accounts.get(&depositor_account_id).unwrap() - amount;
-            self.token.accounts.insert(&depositor_account_id, &new_user_balance);
+            self.burn(&depositor_account_id, amount);
             let mut node = self.get_expect_node_by_ed25519_public_key(ed25519_public_key.clone());
             node.balance += amount;
             let node_account_id = self.nodes_by_ed25519_public_key.get(&ed25519_public_key).unwrap();
@@ -152,8 +151,7 @@ impl MainchainContract {
             assert!(deposited >= amount, "Not enough balance to withdraw");
 
             // subtract from contract balance and add to user balance
-            let new_user_balance = self.token.accounts.get(&depositor_account_id).unwrap() + amount;
-            self.token.accounts.insert(&depositor_account_id, &new_user_balance);
+            self.mint(&depositor_account_id, amount);
             node.balance -= amount;
             let node_account_id = self.nodes_by_ed25519_public_key.get(&ed25519_public_key).unwrap();
             self.handle_node_balance_update(&node_account_id, &node);
--- a/contracts/src/node_registry.rs
+++ b/contracts/src/node_registry.rs
@@ -10,7 +10,7 @@ use near_sdk::{
     Balance,
 };
 
-use crate::{manage_storage_deposit, MainchainContract, MainchainContractExt};
+use crate::{manage_storage_deposit, MainchainContract, MainchainContractExt, MainchainStorageKeys};
 
 /// Deposit info for one account to a node
 #[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize, Eq, PartialEq, Debug, Clone)]
@@ -111,14 +111,9 @@ impl MainchainContract {
 
             // update info for depositor
             let mut depositor = self.depositors.get(&depositor_account_id).unwrap_or_else(|| {
-                // Constructing a unique prefix for a nested UnorderedMap from a concatenation
-                // of a prefix and a hash of the depositor account id.
-                let prefix: Vec<u8> = [
-                    b"d".as_slice(),
-                    &near_sdk::env::sha256_array(depositor_account_id.as_bytes()),
-                ]
-                .concat();
-                UnorderedMap::new(prefix)
+                UnorderedMap::new(MainchainStorageKeys::Depositor {
+                    account_hash: env::sha256_array(depositor_account_id.as_bytes()),
+                })
             });
             depositor.insert(&ed25519_public_key, &amount);
             self.depositors.insert(&depositor_account_id, &depositor);
--- a/contracts/src/node_registry.rs
+++ b/contracts/src/node_registry.rs
@@ -5,26 +5,35 @@ use near_sdk::{
     log,
     near_bindgen,
     serde::{Deserialize, Serialize},
-    Balance, PublicKey,
+    AccountId,
+    Balance,
 };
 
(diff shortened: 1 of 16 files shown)


contracts/src/test_utils.rs
```diff
@@ -29,61 +26,123 @@ pub fn new_contract() -> MainchainContract {
...
 }
 
+pub struct TestAccount {
+    pub account_id:         AccountId,
+    pub ed25519_public_key: near_sdk::PublicKey,
+    pub bn254_private_key:  PrivateKey,
+    pub bn254_public_key:   PublicKey,
+}
+
+pub fn bob() -> TestAccount {
```
gluax, line 36: It looks like a lot of these functions are similar, we could probably boil it down to one function. I.e. `make_test_account(pk: &str, pk_bytes: &[u8], account_id: &str)`.

Then if possible could make `static` instances of the `TestAccount` types. If that can't be done cause certain methods aren't constant we could do the following:

Declare a new type `pub struct TestAccounts;`, then make those `TestAccount` instances associated consts.

Either way cleans up some repetitive code and lets us have constants rather than recreating these every time.

gluax (changes requested):

contracts/src/node_registry.rs
```diff
@@ -80,46 +97,79 @@ impl MainchainContract {
...
         manage_storage_deposit!(self, "require", {
-            let account_id = env::signer_account_id();
+            let depositor_account_id = env::signer_account_id();
 
             // subtract from user balance and add to contract balance
-            let new_user_balance = self.token.accounts.get(&account_id).unwrap() - amount;
-            self.token.accounts.insert(&account_id, &new_user_balance);
-            let mut node = self.get_expect_node(account_id.clone());
+            let new_user_balance = self.token.accounts.get(&depositor_account_id).unwrap() - amount;
+            self.token.accounts.insert(&depositor_account_id, &new_user_balance);
```
FranklinWaller, line 106: These two lines seems like they are going to be used often (also with "transferring" tokens) Might be good to have somewhere an internal util function that handles these things automatically.

contracts/src/node_registry.rs
```diff
@@ -80,46 +97,79 @@ impl MainchainContract {
...
-            self.handle_node_balance_update(&account_id, &node);
+            let node_account_id = self.nodes_by_ed25519_public_key.get(&ed25519_public_key).unwrap();
+            self.handle_node_balance_update(&node_account_id, &node);
+
+            // update info for depositor
+            let mut depositor = self.depositors.get(&depositor_account_id).unwrap_or_else(|| {
+                // Constructing a unique prefix for a nested UnorderedMap from a concatenation
+                // of a prefix and a hash of the depositor account id.
+                let prefix: Vec<u8> = [
+                    b"d".as_slice(),
```
FranklinWaller, line 117: I'm seeing a lot of these prefixes, can we have them in one file containing constants? using just one letter might accidentally collide in the future. Having them in one file gives a clear overview what we are using

contracts/src/node_registry.rs
```diff
@@ -48,8 +57,16 @@ impl MainchainContract {
...
-        self.internal_get_node(&node_id).expect("Node does not exist")
+    pub fn get_expect_node(&self, account_id: AccountId) -> Node {
+        self.internal_get_node(&account_id).expect("Node does not exist")
+    }
+
+    pub fn get_expect_node_by_ed25519_public_key(&self, ed25519_public_key: Vec<u8>) -> Node {
+        let account_id = self
+            .nodes_by_ed25519_public_key
+            .get(&ed25519_public_key)
+            .expect("Node does not exist");
```
FranklinWaller, line 68: Can we make this "Node {ed25519_public_key} does not exist"? It helps a lot

contracts/src/node_registry.rs
jamesondh: Moved this to `MainchainStorageKeys` in `lib.rs` which is a common pattern

FranklinWaller (approved):

contracts/src/node_registry.rs
```diff
@@ -80,46 +99,72 @@ impl MainchainContract {
...
+                .expect("No deposit info found for this account");
+            let deposited = depositor
+                .get(&ed25519_public_key)
+                .expect("No deposit info found for this node");
+            assert!(deposited >= amount, "Not enough balance to withdraw");
+
             // subtract from contract balance and add to user balance
-            let new_user_balance = self.token.accounts.get(&account_id).unwrap() + amount;
-            self.token.accounts.insert(&account_id, &new_user_balance);
+            self.mint(&depositor_account_id, amount);
```
mariocao, line 154: I think that minting/burning for staking/unstaking is weird.
It affects the total token supply and it will make things harder to track (I think).

Question: does it not make more sense to transfer the tokens to the contract?

jamesondh: The FungibleToken standard only exposes `ft_transfer()` which transfers from `env::predecessor_account_id()` to a receiver, so we have to use our own internal transfer function anyways. So actually transferring tokens in `self.tokens` would only add an extra step, since we already track the balance in the `Node` info and globally with `self.last_total_balance`. With this pattern though, it's probably a good idea to override `ft_total_supply()` from the FungibleToken standard and set it to `self.token.total_supply + self.last_total_balance`.

contracts/src/test_utils.rs
jamesondh: I created a `make_test_account` function but not sure if making them static/const is still applicable since I reused Menna's key generation test here

contracts/src/node_registry.rs
FranklinWaller: It's true we track it internally but from an explorer point of view (near blocks for example)it would show those tokens are lost somehow. I think it'd be good to still have in the token mapping show that this account id has the staked tokens

gluax (approved):

contracts/src/test_utils.rs
gluax: Yea, this looks much cleaner to me!

PR #93, opened 2023-03-14
Delegate CLI register, setup & stake

FranklinWaller:
## Motivation

Have a CLI tool where a user can register, stake and top-up a validator node.

## Explanation of Changes

* Add a new project to the workspace "delegate-cli"
* Implement stake `seda-delegate stake <DELEGATION_CONTRACT_ID> <AMOUNT>`
* Implement register `seda-delegate register <DELEGATION_CONTRACT_ID> [MULTI_ADDR]`
* Implement setup `seda-delegate setup <DELEGATION_CONTRACT_ID> [TOPUP_AMOUNT] [STAKE_AMOUNT] [MULTI_ADDR]`

## Testing

Create a `.env` file where the following variables exist:

```
# secret key that is generated by the node itself (seda_key)
VALIDATOR_SECRET_KEY = "FILL_ME_IN"

# The NEAR ed25519 account secret key that holds the SEDA/NEAR tokens
ACCOUNT_SECRET_KEY = "ed25519:FILL_ME_IN"

# The NEAR account id that holds the SEDA/NEAR tokens (and are from the secret key)
SIGNER_ACCOUNT_ID = "FILL_ME_IN.near"
```

Then you can run one of the above mentioned commands

### DIFF
--- a/config/src/configs/delegate.rs
+++ b/config/src/configs/delegate.rs
@@ -28,6 +28,9 @@ pub struct PartialDelegateConfig {
     /// An option to override the RPC URL
     #[arg(long)]
     pub rpc_url:              Option<String>,
+    /// An option to override the node gas config value.
+    #[arg(short, long)]
+    pub gas:                  Option<u64>,
 }
 
 #[cfg(feature = "delegate-cli")]
@@ -49,6 +52,7 @@ impl PartialDelegateConfig {
 
         let delegate_contract_id = merge_config_cli!(self, cli_options, delegate_contract_id, Ok(String::new()))?;
         let rpc_url = merge_config_cli!(self, cli_options, rpc_url, Ok(DelegateConfigInner::RPC_URL.to_string()))?;
+        let gas = merge_config_cli!(self, cli_options, gas, Ok(DelegateConfigInner::GAS))?;
 
         Ok(Arc::new(DelegateConfigInner {
             validator_secret_key,
@@ -56,6 +60,7 @@ impl PartialDelegateConfig {
             signer_account_id,
             delegate_contract_id,
             rpc_url,
+            gas,
         }))
     }
 }
@@ -69,6 +74,7 @@ impl Config for PartialDelegateConfig {
             account_secret_key:   None,
             signer_account_id:    None,
             rpc_url:              None,
+            gas:                  Some(DelegateConfigInner::GAS),
         }
     }
 
@@ -86,6 +92,7 @@ pub struct DelegateConfigInner {
     pub signer_account_id:    String,
     pub delegate_contract_id: String,
     pub rpc_url:              String,
+    pub gas:                  u64,
 }
 
 impl DelegateConfigInner {
@@ -97,6 +104,7 @@ impl DelegateConfigInner {
             validator_secret_key: String::new(),
             account_secret_key:   String::new(),
             signer_account_id:    String::new(),
+            gas:                  Self::GAS,
         })
     }
 
@@ -107,6 +115,7 @@ impl DelegateConfigInner {
 }
 
 impl DelegateConfigInner {
+    pub const GAS: u64 = 300_000_000_000_000;
     pub const RPC_URL: &str = "https://rpc.testnet.near.org";
 }
 
--- a/config/src/configs/delegate.rs
+++ b/config/src/configs/delegate.rs
@@ -115,6 +115,7 @@ impl DelegateConfigInner {
 }
 
 impl DelegateConfigInner {
+    // 300 Tgas
     pub const GAS: u64 = 300_000_000_000_000;
     pub const RPC_URL: &str = "https://rpc.testnet.near.org";
 }
--- a/delegate-cli/Cargo.toml
+++ b/delegate-cli/Cargo.toml
@@ -9,11 +9,16 @@ name = "seda-delegate"
 path = "src/main.rs"
 
 [dependencies]
+bn254 = { workspace = true }
+bs58 = "0.4.0"
 clap = { workspace = true, features = ["default"] }
 clap-markdown = { workspace = true }
 clap_complete = { workspace = true }
 dotenv = { workspace = true }
+hex = "0.4"
 seda-chains = { workspace = true }
-seda-config = { workspace = true, features = ["cli", "delegate-cli"]}
+seda-config = { workspace = true, features = ["cli", "delegate-cli"] }
+seda-crypto = { workspace = true }
 seda-runtime-sdk = { workspace = true }
+serde_json = { workspace = true, features = ["std"] }
 tokio = { workspace = true }
--- a/delegate-cli/Cargo.toml
+++ b/delegate-cli/Cargo.toml
@@ -10,15 +10,16 @@ path = "src/main.rs"
 
 [dependencies]
 bn254 = { workspace = true }
-bs58 = "0.4.0"
+bs58 = { workspace = true }
 clap = { workspace = true, features = ["default"] }
 clap-markdown = { workspace = true }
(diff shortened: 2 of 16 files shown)


delegate-cli/Cargo.toml
```diff
@@ -9,11 +9,16 @@ name = "seda-delegate"
 path = "src/main.rs"
 
 [dependencies]
+bn254 = { workspace = true }
+bs58 = "0.4.0"
 clap = { workspace = true, features = ["default"] }
 clap-markdown = { workspace = true }
 clap_complete = { workspace = true }
 dotenv = { workspace = true }
+hex = "0.4"
```
gluax, line 18: `workspace = true`?

delegate-cli/Cargo.toml
```diff
@@ -9,11 +9,16 @@ name = "seda-delegate"
 path = "src/main.rs"
 
 [dependencies]
+bn254 = { workspace = true }
+bs58 = "0.4.0"
```
gluax, line 13: `workspace = true`?

delegate-cli/src/cli/commands/register.rs
```diff
@@ -0,0 +1,61 @@
...
+pub struct Register {
+    /// The contract address to register on
+    pub delegation_contract_id: String,
+    /// The multi address that is associated with the node
+    pub multi_addr:             Option<String>,
+}
+
+impl Register {
+    pub async fn handle(self, config: DelegateConfig) {
+        let bn254_key = derive_bn254_key_pair(&config.validator_secret_key, 0).unwrap();
```
gluax, line 21: Should add `thiserror` and define an `error` type. So we could avoid all these `unwrap` calls.

delegate-cli/src/cli/commands/register.rs
```diff
@@ -0,0 +1,61 @@
...
+        let ed25519_key = derive_ed25519_key_pair(&config.validator_secret_key, 0).unwrap();
+        let ed25519_public_key = ed25519_key.public_key.as_bytes().to_vec();
+        let account_id = hex::encode(&ed25519_public_key);
+        let signature = ECDSA::sign(&account_id, &bn254_key.private_key).unwrap();
+        let ed25519_secret_key_bytes: Vec<u8> = ed25519_key.into();
+
+        // TODO: Make construct_signed_tx only accept bytes and not strings
+        let signed_tx = chain::construct_signed_tx(
```
gluax, line 29: I think we have functionality for making these calls easier from the CLI in the existing CLI crate. You can just slap a `TODO` here for now and we can explore that in a follow-up PR since it would require some moving stuff around.

delegate-cli/src/cli/commands/register.rs
```diff
@@ -0,0 +1,61 @@
...
+use clap::Args;
+use seda_chains::{chain, Client};
+use seda_config::{ChainConfigsInner, DelegateConfig};
+use seda_crypto::{derive_bn254_key_pair, derive_ed25519_key_pair};
+use seda_runtime_sdk::Chain;
+use serde_json::json;
+
+use crate::cli::utils::to_yocto;
+
+#[derive(Debug, Args)]
```
gluax, line 11: We should have `#[clap(short, long)]` above these fields. So, users can enter data in whatever order with the arg flags. But that's just my take.

delegate-cli/src/cli/commands/setup.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub multi_addr:             Option<String>,
+}
+
+impl Setup {
+    pub async fn handle(self, config: DelegateConfig) {
+        let ed25519_key = derive_ed25519_key_pair(&config.validator_secret_key, 0).unwrap();
+        let ed25519_public_key = ed25519_key.public_key.as_bytes().to_vec();
+
+        let top_up = TopUp {
+            amount:   self.topup_amount.unwrap_or("5".to_string()),
```
gluax, line 25: You can set default values with `clap`. Similar to the above comment except this time it's a String.

delegate-cli/src/cli/commands/register.rs
```diff
@@ -0,0 +1,61 @@
...
+
+        // TODO: Make construct_signed_tx only accept bytes and not strings
+        let signed_tx = chain::construct_signed_tx(
+            Chain::Near,
+            &account_id,
+            &bs58::encode(ed25519_secret_key_bytes).into_string(),
+            &self.delegation_contract_id,
+            "register_node",
+            json!({
+                "multi_addr": self.multi_addr.unwrap_or(String::new()),
```
gluax, line 36: You can set defaults with `clap`. i.e. make the `multi_addr` a `String` instead of `Option<String>` and then tell clap to do a default value of empty string.

delegate-cli/src/cli/commands/setup.rs
```diff
@@ -0,0 +1,45 @@
+use clap::Args;
+use seda_config::DelegateConfig;
+use seda_crypto::derive_ed25519_key_pair;
+
+use super::{register::Register, stake::Stake, top_up::TopUp};
+
+#[derive(Debug, Args)]
+pub struct Setup {
```
gluax, line 8: I do think the `amount` fields should be numbers, so the validation is done for us by `clap` who prints nice user errors :).

delegate-cli/src/cli/commands/stake.rs
```diff
@@ -0,0 +1,58 @@
...
+use serde_json::json;
+
+use crate::cli::utils::to_yocto;
+
+#[derive(Debug, Args)]
+pub struct Stake {
+    /// The contract address to stake on
+    pub delegation_contract_id: String,
+    /// The amount of SEDA tokens to stake (1 = 1 SEDA)
+    pub amount:                 String,
```
gluax, line 15: Same here, a number type rather than a String. We can always call `to_string` on the number.

gluax (dismissed):

delegate-cli/src/cli/commands/register.rs
FranklinWaller: I think with these amount of arguments it's not needed, with setup it might be useful

Thomasvdam (approved):
From a super outsider perspective it at least seems clear what is doing what. Can't really say much about correctness or code style though 😬

delegate-cli/src/cli/commands/register.rs
```diff
@@ -0,0 +1,65 @@
...
+use serde_json::json;
+
+use crate::cli::{errors::Result, utils::to_yocto};
+
+#[derive(Debug, Args)]
+pub struct Register {
+    /// The contract address to register on
+    pub delegation_contract_id: String,
+
+    #[clap(default_value = "")]
```
mariocao, line 16: Maybe stupid question, was not the default multi address better something like `/ip4/0.0.0.0/tcp/0`?

config/src/configs/delegate.rs
```diff
@@ -107,6 +115,7 @@ impl DelegateConfigInner {
 }
 
 impl DelegateConfigInner {
+    pub const GAS: u64 = 300_000_000_000_000;
```
mariocao, line 118: In reality this amount is the maximum of gas units that can be set in a transaction.
```suggestion
    // 300 Tgas
    pub const GAS: u64 = 300_000_000_000_000;
```

mariocao (approved):

delegate-cli/src/cli/commands/register.rs
FranklinWaller: Yeah would go for the second option. The /ip4/0.0.0.0/tcp/0 stores unnecessary information on the main chain

PR #94, opened 2023-03-15
Feat/common module

gluax:
## Motivation

To have commonly used Data Structures from a single source rather than re-defining them.

## Explanation of Changes

This creates a new module within our workspace, `seda-common`, and we define the commonly used structures there.
This module exposes a `cli` flag that provides clap bindings.

The `contracts` and `cli` modules now use these as a single source of truth. 

The update command has been changed to be a struct enum in a single place so that it also works for the CLI.

It also adds a `test` feature to the module so that we can generate better mock data for our mock rpc.

**NOTES**: 

- ~~The register command needs the private key and signature. I have dummy values there since we don't store the `KeyPair` on the  `NodeConfig` in this branch. Once that change is in, we can address that.~~
- Similarly, the `unregister` command has been removed from the contract but is to be added back in. So, for now, I'm not making any changes to it, and when we add that command back in, we can fix it.

## Testing

By hand.

## Related PRs and Issues

~~#85 adds the `KeyPair` in the `NodeConfig`, so that PR can make the changes to fix the `register` command if this one is merged first. Or we can do it in a separate follow-up PR. ~~
~~#91 Has changes made to some of these structs in its PR, so there will be conflicts one way or another after one is merged.~~


seda_sk
jamesondh, line 1: should be .gitignored?

jamesondh (changes requested):

seda_sk
FranklinWaller: Did a secret key just got leaked 😛

common/src/node/get.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl ToString for GetNodeArgs {
+    fn to_string(&self) -> String {
+        let json = json!(self);
+        json.to_string()
+    }
+}
+
+#[derive(Debug, Clone, Default, Eq, PartialEq, BorshDeserialize, BorshSerialize, Deserialize, Serialize)]
+
```
mariocao, line 22:
```suggestion
```

common/src/node/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+mod register;
+pub use register::*;
+
+mod update;
+pub use update::*;
+
+use super::*;
+
+#[derive(Debug, Clone, Eq, PartialEq, BorshDeserialize, BorshSerialize, Deserialize, Serialize)]
+pub struct HumanReadableNode {
```
mariocao, line 13: Maybe `NodeInfo`?
```suggestion
pub struct NodeInfo {
```

seda_sk
gluax: whoops! it was ignored in my other branch but not this one 😅

jamesondh (approved):

FranklinWaller (approved):

PR #95, opened 2023-03-16
Feat/slot leader

mennatabuelnaga:
- Adds a function to calculate the current slot leader: `get_current_slot_leader`
- Adds a global variable `last_generated_random_number` instantiated at contract creation via a parameter.

### DIFF
--- a/contracts/src/committee.rs
+++ b/contracts/src/committee.rs
@@ -38,6 +38,16 @@ impl MainchainContract {
         }
         chosen_committee
     }
+
+    pub fn select_slot_leader(&self, random_number: u64) -> AccountId {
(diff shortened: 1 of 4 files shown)


contracts/src/committee.rs
```diff
@@ -46,6 +49,14 @@ impl MainchainContract {
     pub fn get_committees(&self) -> Vec<Vec<AccountId>> {
         self.committees.clone()
     }
```
gluax, line 51: Add a new line between functions?

gluax (approved):

Thomasvdam (approved):
Can't say too much about correctness/style, but it is clear to me how the code is intended to work :)

jamesondh (approved):

PR #97, opened 2023-03-17
Feat/contract withdraw restrictions

jamesondh:
## Motivation

Similar to how nodes should only be activated after a waiting period, nodes should only be able to withdraw after a waiting period.

## Explanation of Changes

- Added `request_withdraw()` function that allows user to withdraw given amount after x epochs. Only one withdraw request is allowed at one time
- Added `cancel_withdraw_request()` to remove previously created withdraw request
- Added `unregister_node()` function that allows user to return storage deposit and delete the node when balance is 0

## Testing

- cancel_withdraw_request() - tests `cancel_withdraw_request()` deletes a pending request
- withdraw_before_epoch() - tests that user cannot immediately withdraw funds after creating withdraw request
- unregister_nonzero_node() - tests that user cannot unregister node if there is balance remaining

### DIFF
--- a/contracts/src/node_registry.rs
+++ b/contracts/src/node_registry.rs
@@ -246,6 +246,22 @@ impl MainchainContract {
         });
     }
 
+    pub fn unregister_node(&mut self, ed25519_public_key: Vec<u8>) {
+        // assert the signer_account_pk matches the ed25519_public_key
+        assert!(
+            env::signer_account_pk().into_bytes().to_vec() == ed25519_public_key,
+            "Invalid ed25519_public_key"
+        );
+
+        // assert the node balance is zero
+        let node = self.get_expect_node_by_ed25519_public_key(ed25519_public_key.clone());
+        assert!(node.balance == 0, "Node balance is not zero");
+
+        // remove the node
+        let account_id = self.nodes_by_ed25519_public_key.get(&ed25519_public_key).unwrap();
+        self.inactive_nodes.remove(&account_id);
+    }
+
     /// Updates one of the node's fields
     #[payable]
     pub fn update_node(&mut self, command: UpdateNode) {
--- a/contracts/src/node_registry.rs
+++ b/contracts/src/node_registry.rs
@@ -1,6 +1,6 @@
 use near_sdk::{
     borsh::{self, BorshDeserialize, BorshSerialize},
-    collections::UnorderedMap,
+    collections::{LookupMap, UnorderedMap},
     env,
     json_types::{U128, U64},
     log,
@@ -17,6 +17,13 @@ use crate::{manage_storage_deposit, MainchainContract, MainchainContractExt, Mai
 pub struct HumanReadableDepositInfo {
     pub node_ed25519_public_key: Vec<u8>,
     pub amount:                  Balance,
+}
+
+/// Withdraw request info for one account to a node
+#[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize, Eq, PartialEq, Debug, Clone)]
+pub struct WithdrawRequest {
+    pub amount: Balance,
+    pub epoch:  u64,
 }
 /// Node information
 #[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize, Eq, PartialEq, Debug, Clone, Default)]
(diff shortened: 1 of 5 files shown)


gluax (approved):

contracts/src/node_registry.rs
```diff
@@ -134,7 +141,7 @@ impl MainchainContract {
 
     pub fn internal_withdraw(&mut self, amount: Balance, ed25519_public_key: Vec<u8>) {
         // TODO: epoch delay for withdrawal
```
Thomasvdam, line 143: Doesn't this PR implement this TODO or am I missing something?

contracts/src/node_registry.rs
```diff
@@ -150,6 +157,23 @@ impl MainchainContract {
...
+            });
+            // assert there is a pending withdrawal for this depositor
+            let withdraw_request = node_withdraw_requests
+                .get(&depositor_account_id)
+                .expect("No pending withdrawal request found for this account");
+            // check that the epoch is valid
+            let current_epoch = self.get_current_epoch();
+            assert!(
+                withdraw_request.epoch <= current_epoch,
+                "Not enough epochs have passed to withdraw"
```
Thomasvdam, line 174: Perhaps a QoL addition would be to mention how many more epochs remain before they can withdraw?

contracts/src/node_registry.rs
```diff
@@ -161,6 +185,11 @@ impl MainchainContract {
             // update global balance
             self.last_total_balance -= amount;
 
+            // remove the withdraw request
+            node_withdraw_requests.remove(&depositor_account_id);
+            self.withdraw_requests
+                .insert(&ed25519_public_key, &node_withdraw_requests);
+
```
Thomasvdam, line 192: Just a check question as I'm not yet well versed enough in smart contracts and their flows: this is not susceptible to a reentrancy attack where an attacker can 'claim' the withdrawal multiple times before it gets removed from our withdrawal requests?

contracts/src/node_registry.rs
```diff
@@ -246,6 +293,82 @@ impl MainchainContract {
         self.internal_deposit(amount, ed25519_public_key);
     }
 
+    #[payable]
+    pub fn request_withdraw(&mut self, amount: U128, ed25519_public_key: Vec<u8>) {
+        manage_storage_deposit!(self, "require", {
+            let amount: Balance = amount.into();
+            assert!(amount > 0, "Withdrawal amount should be positive");
```
Thomasvdam, line 300: Does it make sense to allow withdrawing 0?

contracts/src/node_registry.rs
```diff
@@ -246,6 +293,82 @@ impl MainchainContract {
...
+                    account_hash: env::sha256_array(ed25519_public_key.as_slice()),
+                })
+            });
+            let withdraw_request = node_withdraw_requests
+                .get(&depositor_account_id)
+                .expect("No pending withdrawal found for this account");
+            assert!(
+                withdraw_request.epoch <= self.get_current_epoch(),
+                "Not enough epochs have passed to cancel the withdrawal request"
+            );
```
Thomasvdam, line 359: Maybe this is normal in crypto land, but it feels odd you're unable to cancel a request before it would be open to actually withdraw? Is this done to prevent abuse?

mariocao: maybe I am missing something but why would we consider this flow?

contracts/src/node_registry.rs
```diff
@@ -18,6 +18,13 @@ pub struct HumanReadableDepositInfo {
     pub node_ed25519_public_key: Vec<u8>,
     pub amount:                  Balance,
 }
+
+/// Withdraw request info for one account to a node
+#[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize, Eq, PartialEq, Debug, Clone)]
+pub struct WithdrawRequest {
+    pub amount: Balance,
+    pub epoch:  u64,
```
mariocao, line 26: I would either comment what is this epoch (e.g., epoch after which funds can be withdrawn) or I would rename it to something like `activation_epoch`.

contracts/src/node_registry.rs
jamesondh: Good question! Reentrancy attacks are only possible in NEAR with cross-contract calls, which we avoided by merging the mainchain and token contracts. This is because calls are processed as promises, and cross-contract calls are asynchronous and happen over multiple blocks.

(difficult to find a primary source/documentation on this, but [others have talked about it elsewhere](https://twitter.com/moo9000/status/1321031995154141185))

contracts/src/node_registry.rs
jamesondh: I can't think of a compelling reason either, so I removed this, thanks for pointing it out!

contracts/src/node_registry.rs
Thomasvdam: Thanks for the explanation and the link, good resources 😃

Thomasvdam (approved):

PR #101, opened 2023-03-22
Ref/remove private keys from wasm

FranklinWaller:
## Motivation

(Co-Authored by @mariocao)

This PR includes several changes:
* Introduce new MasterKey struct which is used as the base to derive to bn254 and ed25519 keys. 
* Remove any references of private keys inside the wasm binary

## Explanation of Changes

* Remove node config from wasm binary, signing now takes places in the imports
* create new MasterKey struct which is used to generate random keys/write/derive


crypto/src/test.rs
Thomasvdam, line 1: Thanks to Mario I understand that the `index` param is to support key rotation. Perhaps it's an idea to add a test that verifies you actually get a different keypair when you change the `index` value?

Thomasvdam (approved):
More of a 'looks good to me' kind of approve rather than an "I'm confident this works perfectly", so keep that in mind. :D

runtime/core/src/runtime_test.rs
```diff
@@ -55,6 +62,7 @@ async fn test_promise_queue_multiple_calls_with_external_traits() {
     );
 
     let vm_result = runtime_execution_result.await;
+    dbg!(&vm_result);
```
gluax, line 65: oops.

wasm/consensus/src/tasks/bridge.rs
```diff
@@ -46,7 +46,7 @@ fn bridge_step_1() {
             log!(Level::Debug, "Posting args: {args_string}");
             chain_call(
                 Chain::Near,
-                CONFIG.contract_account_id.as_str(), // TODO: Currently panics
+                get_oracle_contract_id().as_str(), // TODO: Currently panics
```
gluax, line 49: Maybe for now we can do `memory_read` and `memory_write` again and leave the TODO above it? So we can test it all still works manually with the `bridge` debug cli command.

mariocao:
Superseded by PR #110

PR #106, opened 2023-03-28
fix(contracts): filter inactive_nodes in process_epoch()

mennatabuelnaga:
- Filters inactive_nodes in process_epoch.
- Adds get_inactive_nodes()
- renames get_nodes() to get_active_nodes()
- Tests filtered inactive_nodes within the get_nodes() test.

### DIFF
--- a/contracts/src/epoch.rs
+++ b/contracts/src/epoch.rs
@@ -30,6 +30,7 @@ impl MainchainContract {
                 if activation_epoch <= &self.get_current_epoch() {
                     self.active_nodes
                         .insert(account_id, &self.inactive_nodes.get(account_id).unwrap());
+                    self.inactive_nodes.remove(account_id);
                     // log!("Moving pending node {} to active nodes", account_id);
                     false
                 } else {
--- a/contracts/src/node_registry.rs
+++ b/contracts/src/node_registry.rs
@@ -181,6 +181,32 @@ impl MainchainContract {
             );
         });
     }
+
+    pub fn internal_get_nodes(
+        &self,
+        nodes_map: &UnorderedMap<AccountId, Node>,
+        limit: U64,
+        offset: U64,
+    ) -> Vec<NodeInfo> {
+        let mut nodes = Vec::new();
+        let mut index = nodes_map.len() - u64::from(offset);
+        let limit = u64::from(limit);
+        while index > 0 && nodes.len() < limit.try_into().unwrap() {
+            if let Some(node_id) = nodes_map.keys().nth(index as usize - 1) {
+                let node = nodes_map.get(&node_id).unwrap();
+                let human_readable_node = NodeInfo {
+                    account_id:         node_id.to_string(),
+                    multi_addr:         node.multi_addr,
+                    balance:            node.balance,
+                    ed25519_public_key: node.ed25519_public_key,
+                    bn254_public_key:   node.bn254_public_key,
+                };
+                nodes.push(human_readable_node);
+            }
+            index -= 1;
+        }
+        nodes
+    }
 }
 
 /// Contract public methods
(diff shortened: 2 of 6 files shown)


contracts/src/epoch.rs
```diff
@@ -30,6 +30,7 @@ impl MainchainContract {
                 if activation_epoch <= &self.get_current_epoch() {
                     self.active_nodes
                         .insert(account_id, &self.inactive_nodes.get(account_id).unwrap());
+                    self.inactive_nodes.remove(account_id);
```
Thomasvdam, line 33: If I understand the docs correctly doesn't remove return the value if it was in the map? https://docs.rs/near-sdk/latest/near_sdk/collections/unordered_map/struct.UnorderedMap.html#method.remove

If so I'd change the code to be something like
```rust
let node = self.inactive_nodes.remove(account_id);
// Some check to see if you actually got a node, don't know the syntax yet 😬 
// I guess it's some magic if let Ok()
if check_I_need_to_learn {
  self.active_nodes.insert(account_id, &node);
}
```

Thomasvdam (approved):

contracts/src/epoch.rs
jamesondh: If we weren't certain that the node was a member of `inactive_nodes` this check would make sense, but I think it's reasonable to assume so since the `handle_node_balance_update()` function adds the node to `inactive_nodes` when also adding to `pending_nodes`: https://github.com/sedaprotocol/seda-rust/blob/b5b49987895570ca9f5a5eb7b9fb5dd12f2e20d0/contracts/src/node_registry.rs#L42

jamesondh (changes requested):
Lgtm but would add a test as Thomas mentioned to verify the updated behavior works as intended

contracts/src/node_registry.rs
```diff
@@ -383,6 +383,27 @@ impl MainchainContract {
...
+                    balance:            node.balance,
+                    ed25519_public_key: node.ed25519_public_key,
+                    bn254_public_key:   node.bn254_public_key,
+                };
+                nodes.push(human_readable_node);
+            }
+            index -= 1;
+        }
+        nodes
+    }
```
Thomasvdam, line 405: This looks similar to the `get_active_nodes` one. Is it possible to do something like:

```rust
fn get_nodes(nodesMap: UnorderedMap<AccountId, Node>, limit: U64, offset: U64) -> Vec<NodeInfo> {
        let mut nodes = Vec::new();
        let mut index = nodesMap.len() - u64::from(offset);
        let limit = u64::from(limit);
        while index > 0 && nodes.len() < limit.try_into().unwrap() {
            if let Some(node_id) = nodesMap.keys().nth(index as usize - 1) {
                let node = nodesMap.get(&node_id).unwrap();
                let human_readable_node = NodeInfo {
                    account_id:         node_id.to_string(),
                    multi_addr:         node.multi_addr,
                    balance:            node.balance,
                    ed25519_public_key: node.ed25519_public_key,
                    bn254_public_key:   node.bn254_public_key,
                };
                nodes.push(human_readable_node);
            }
            index -= 1;
        }
        nodes
    }

pub fn get_inactive_nodes(&self, limit: U64, offset: U64) -> Vec<NodeInfo> {
// Probably need some other magic lifetime symbol but I hope it gets the point across
  self.get_nodes(&self.inactive_nodes, limit, offset);
}
```

mennatabuelnaga: That's actually a better way to do it, thank you!

jamesondh (changes requested):
`get_nodes()` is referenced in the CLI and test RPC, so we should either keep it on the contract (maybe combining `get_active_nodes()` and `get_inactive_nodes()`) or update it in those two places in the repo.

contracts/src/node_registry.rs
```diff
@@ -160,6 +160,32 @@ impl MainchainContract {
             );
         });
     }
+
+    pub fn internal_get_nodes(
```
Thomasvdam, line 164: Don't know enough about rust, does this need to be public? 😬

Thomasvdam (approved):

jamesondh (approved):

PR #107, opened 2023-03-28
Call `process_epoch()` before signed batch submission in first slot

jamesondh:
## Motivation

Instead of needing to manually call process_epoch() to select committees, activate pending nodes, etc, this call should be included in post_signed_batch() to reduce complexity from the node operator’s side.

## Explanation of Changes

- `post_signed_batch()` always calls `process_epoch()`, which will select committees for an epoch in advance as long as there is a new epoch that has not yet been processed.
- Refactored `get_committees()` into `get_committee(epoch_height)` to make logic cleaner and avoid accidentally using the wrong committee for an epoch. Also changed the `committees` state variable from `Vec<Vec<AccountId>>` to `LookupMap<EpochHeight, Vec<AccountId>>`

## Testing

`integration_test_1()` checks if an epoch unexpectedly changes. `process_epoch()` is only called to bootstrap the committees and then lets `post_signed_batch()` handle the rest of the epoch processing.

## Related PRs and Issues

N/A


contracts/src/integration_test.rs
```diff
@@ -104,6 +104,15 @@ fn integration_test_1() {
...
 
+            // if we are in the last slot of this epoch, check the logs to see if
+            // process_epoch() was called
+            if contract.get_current_slot_within_epoch() == SLOTS_PER_EPOCH - 1 {
+                assert_eq!(
+                    get_logs().first().unwrap(),
+                    &"Posted batch for last slot within epoch, calling `process_epoch()`".to_string()
+                );
+            }
+
```
mariocao, line 115: I don't understand this part.

The specification for this part described that a new epoch would be processed in the first batch of the epoch.
The underlying reason is that the slots are bounded to NEAR blocks and we cannot guarantee that a slot leader will be able to submit a batch during that time.

Therefore, should process an epoch in the first batch submitted within the new epoch. We are not computing the committee for the current epoch but for a future one. We always know 2 epochs in advance.

mariocao, line 115: This assertion is only verified sometimes, right?
Is it any way to guarantee that `process epoch` was called after an epoch change?

mariocao (approved):

PR #109, opened 2023-03-29
Feat/deploy and setup contracts script

jamesondh:
TODO: script works as-is, but rewriting the deployment/initialization/test account creation in Rust to make CI possible.

## Motivation

(Write your motivation here)

## Explanation of Changes

(Write your explanation here)

## Testing

(Write your test plan here)

## Related PRs and Issues

(Link your related PRs and Issues here)


gluax:
This is still in draft mode, but I have a few suggestions.

1. I'd look into `getopts`, it's built-in to bash, for argument parsing from bash. It will make the script and argument parsing part easier to understand. Here's a helpful [link](https://stackoverflow.com/questions/16483119/an-example-of-how-to-use-getopts-in-bash).
2. This whole `scripts` folder should be moved to `tools/scripts`. Better to have a centralized place for all tooling/scripts we write to help us test. Excluding the `Makefile` which makes sense to always have in the top directory.

PR #110, opened 2023-03-30
feat(wasm): add consensus logic

mariocao:
PR co-authored by @FranklinWaller 

## Motivation

This PR adds functionality for consensus but also fixes some problems that were identified during the implementation.

## Explanation of Changes

This PR includes the first basic consensus logic for batch and p2p handling among other things that were required in the way:
- Refactor regarding `KeyPair` and how they were configured and used.
- Implement batch consensus logic
- Implement slot leader logic
- Implement p2p handler for receiving batch messages

## Testing

Unfortunately, right now there is no automated integration or end-to-end tests. Therefore, this PR has been manually tested.

Testing instructions using `near-cli` (Javscript version, not rust):

### 1. Build contracts

Before building the contracts, we have to comment out the following verifications in `contracts/src/batch.rs`:

Assertion for disallowing empty batches:
```
// require the data request accumulator to be non-empty
// assert!(
//     !self.data_request_accumulator.is_empty(),
//     "Data request accumulator is empty"
// );
```

Verification of aggregated public key using external `bn254` library instead of provided host functions (subsidized):
```
// let aggregate_public_key_reconstructed = signers.iter().skip(1).fold(
//     PublicKey::from_compressed(self.active_nodes.get(&signers[0]).unwrap().bn254_public_key).unwrap(),
//     |acc, signer| {
//         assert!(self.committees[0].contains(signer), "Node is not part of the committee");
//         let signer_public_key =
//             PublicKey::from_compressed(self.active_nodes.get(signer).unwrap().bn254_public_key).unwrap();
//         acc + signer_public_key
//     },
// );
// assert!(
//     aggregate_public_key_reconstructed.to_compressed().unwrap() == aggregate_public_key,
//     "Invalid aggregate public key"
// );
```

This needs to be commented out because otherwise the transaction runs out of gas (>300 Tgas).

```
make build-contracts
```

### 2. Deploy contract using `near-cli` Javascript command

It will deploy the contract to testnet and with a automatically generated account id.

```
near dev-deploy target/wasm32-unknown-unknown/release/seda_mainchain.wasm
```

### 3. Initialize smart contract

Initialize the contracts by calling the `init()` function. For example, it can be done by using the same development account from previous step.

```
near call <dev_account_id> new '{"dao":"<dev_account_id>","initial_supply":"1000000000000000000000000000","metadata":{"spec":"ft-1.0.0","name":"SEDA Token","symbol":"SEDA","icon":null,"reference":null,"reference_hash":null,"decimals":24},"committee_size":'2',"random_seed":"531540567384302522"}' --account-id dev_account_id
```

### 4. Setup node operators

Using the all-in-one CLI command `delegate setup`, new nodes can be configured as:

```
SIGNER_ACCOUNT_ID=<my-account.tesnet> ACCOUNT_SECRET_KEY=<secret_key> VALIDATOR_SECRET_KEY=<my_secret_key> make run-build-delegate setup -- <contract_account_id>
```

### 5. Start the first epoch in the contract

This is done by calling `process_epoch()` function in the contract.

```
near call <contract_id> process_epoch --account-id <account_id>
```

In this case because we are using a dev account, we can set `account_id` as the `contract_id`.

### 6. Run multiple nodes

Run multiple nodes to have a working committee (in my tests there were 3).

For example for node 1:
```
SEDA_SECRET_KEY=<KEY_NODE_1> make run-build-all run -- --p2p-server-address /ip4/127.0.0.1/tcp/50001
```

The config file `config.toml` should be properly configured to use the `contract_id` we just deployed.

### 7. (Optional) Add peers manually

Sometimes, in my local environment peer discovery was not finding the new peers using mDNS. Therefore, I had to add the peers manually.

For example to force the discovery of 3 nodes with consecutive TCP ports:
```
target/debug/seda node peers add /ip4/127.0.0.1/tcp/50001
target/debug/seda node peers add /ip4/127.0.0.1/tcp/50002
target/debug/seda node peers add /ip4/127.0.0.1/tcp/50003
```


chains/src/near_chain.rs
```diff
@@ -117,15 +113,15 @@ impl ChainAdapterTrait for NearChain {
     }
 
     async fn construct_transfer_tx(
-        signer_acc_str: &str,
-        signer_sk_str: &str,
+        signer_account_id: Option<&str>,
+        signer_keypair: Vec<u8>,
```
gluax, line 117: and here as well.

cli/src/cli/commands/node/get/node.rs
```diff
@@ -5,27 +5,39 @@ use seda_runtime_sdk::Chain;
...
-        let args = GetNodeArgs::from(signer_account_id).to_string();
-        view::<Option<NodeInfo>>(
-            Chain::Near,
-            &contract_account_id,
-            "get_node",
-            Some(args),
-            &chains_config,
-        )
-        .await
+        let contract_id = if let Some(contract_id) = self.contract_id {
```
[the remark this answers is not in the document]
gluax: We had something like `to_contract_id` on the config before. Now we could define a `PartialNodeConfig` and say `.to_partial_node_config` given the user config and inputs, which we already have here.

runtime/core/src/storage/in_memory_adapter.rs
```diff
@@ -22,4 +22,8 @@ impl MemoryAdapter for InMemory {
     {
         self.memory.insert(key.into(), value.to_bytes())
     }
+
+    fn contains_key(&self, key: &str) -> bool {
```
gluax, line 26: Do we need this when `get` already returns an `Option`? Or maybe it's to check you aren't overwriting a field?

runtime/sdk/src/wasm/memory.rs
```diff
@@ -43,3 +43,16 @@ pub fn shared_memory_set(key: &str, mut value: Vec<u8>) {
...
+pub fn shared_memory_contains_key(key: &str) -> bool {
+    let key_len = key.len() as i64;
+    let mut key = key.to_string().into_bytes();
+
+    let result = unsafe { raw::shared_memory_contains_key(key.as_mut_ptr(), key_len) };
+
+    match result {
+        0 => false,
+        1 => true,
+        _ => panic!("Bn254 verify returned invalid bool in u8: {}", result),
```
gluax, line 56: Better to use `unreachable!(...)` here instead of `panic!(...)`. Would provide a better message that something *really* went wrong. Mostly because as far as I can tell it should actually be unreachable.

wasm/consensus/src/tasks/batch.rs
```diff
@@ -1,73 +1,273 @@
 use clap::Args;
+use primitive_types::U256;
+use seda_common::{ComputeMerkleRootResult, MainChainConfig};
 use seda_runtime_sdk::{
```
gluax, line 4: maybe with all these imports at this point we should do `use seda_runtime_sdk::{*, wasm::*}`.

wasm/consensus/src/tasks/batch.rs
```diff
@@ -1,73 +1,273 @@
...
+            "get_last_generated_random_number",
+            Vec::new(),
+        ))
         .then(call_self("batch_step_1", vec![]));
     }
 }
 
 #[no_mangle]
 fn batch_step_1() {
-    log!(Level::Debug, "Batch Step 1");
```
gluax, line 43: I personally think we should keep some debug logs here. So we know we hit this function last if that makes sense. Users won't get debug logs in, so there's only benefits to having more for debug purposes imo.

wasm/consensus/src/tasks/batch.rs
```diff
@@ -1,73 +1,273 @@
...
+    let ed25519_public_key = shared_memory_get("ed25519_public_key");
+    let mut signature_store = get_or_create_batch_signature_store(BATCH_SIGNATURE_STORE_KEY);
+
+    // Retrieve results from promise results
+    let batch: ComputeMerkleRootResult = match Promise::result(0) {
+        PromiseStatus::Fulfilled(Some(batch_bytes)) => {
+            let batch_string =
+                String::from_bytes_vec(batch_bytes).expect("Cannot convert `merkle_root` bytes to json string");
+
+            serde_json::from_str(&batch_string).expect("Cannot convert `merkle_root` json to `ComputeMerkleRootResult`")
```
gluax, line 95: Instead of converting to a `String` then doing `serde_json::from_str` you can do `serde_json::from_slice`. Saves a computation step.

wasm/consensus/src/tasks/batch.rs
```diff
@@ -1,73 +1,273 @@
...
+        &batch.current_slot,
+        hex::encode(&batch.merkle_root),
+        Some(&node_implicit_account) == batch.current_slot_leader.as_ref()
+    );
+
+    // ==========================
+    // || Process BATCH Logic  ||
+    // ==========================
+
+    // Case 1. Check if it was already processed
```
gluax, line 132: I think it would make for cleaner code to move this logic to the `batch` type itself(in a method) and pass whatever data it additionally needs.

wasm/consensus/src/tasks/batch.rs
```diff
@@ -1,73 +1,273 @@
...
+        );
+
+        p2p_broadcast_message(signature_store.p2p_message.clone()).start();
+    }
+
+    // ========================
+    // || SLOT LEADER Logic  ||
+    // ========================
+
+    let current_slot_leader = match batch.current_slot_leader {
```
gluax, line 217: Same thing here, should be a method on the type, for cleanliness imo.

wasm/consensus/src/tasks/bridge.rs
```diff
@@ -46,7 +46,7 @@ fn bridge_step_1() {
             log!(Level::Debug, "Posting args: {args_string}");
             chain_call(
                 Chain::Near,
-                CONFIG.contract_account_id.as_str(), // TODO: Currently panics
+                get_oracle_contract_id().as_str(), // TODO: Currently panics
```
gluax, line 49: Can we use memory read on this value in the first step with a `TODO` above it? I think better to have functioning logic. Especially since we have a CLI command to debug this logic.

wasm/consensus/src/tasks/p2p.rs
```diff
@@ -1,13 +1,124 @@
...
+    ///     * Otherwise, it is ignored and P2P counter-measure is triggered
+    ///  2. Check batch header (i.e., merkle root)
+    ///     * If current known batch, append to list of accumulated signatures
+    ///     * If unknown batch, add to future batch list (current batch was
+    ///       submitted)
     pub fn handle(self) {
-        p2p_broadcast_message(self.message.into_bytes()).start();
+        let message = Message::from_str(&self.message).expect("Failed to decode message");
+
+        match message {
```
gluax, line 40: Fine for now, but this logic should move to a similar pattern like our `task.handle` so as we add more in the future the function logic doesn't get too cluttered.

wasm/consensus/src/types/batch_signature.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use seda_runtime_sdk::{
+    wasm::{shared_memory_contains_key, shared_memory_get, Bn254PublicKey, Bn254Signature},
+    FromBytes,
+};
+use serde::{Deserialize, Serialize};
+
+pub const BATCH_SIGNATURE_STORE_KEY: &str = "batch_signatures";
+
+pub static EMPTY_SHA256: [u8; 32] = [
```
gluax, line 11: Sorry I don't remember but does not allow us to make static `Vec` types?

wasm/consensus/src/types/batch_signature.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+}
+
+pub fn get_or_create_batch_signature_store(storage_key: &str) -> BatchSignatureStore {
+    if shared_memory_contains_key(storage_key) {
+        let result = shared_memory_get(storage_key);
+        let json_str = String::from_bytes_vec(result).unwrap();
+
+        serde_json::from_str(&json_str).expect("Invalid stored `BatchSignatureStore` object")
```
gluax, line 62: Same here could use `serde_json::from_slice` and skip the `String` step.

wasm/consensus/src/message.rs
```diff
@@ -0,0 +1,36 @@
+use std::str::FromStr;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BatchMessage {
```
gluax, line 6: i.e. that logic before should be moved to a method on this type.

wasm/consensus/src/message.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub batch_header:       Vec<u8>,
+    pub bn254_public_key:   Vec<u8>,
+    pub signature:          Vec<u8>,
+    pub ed25519_public_key: Vec<u8>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Message {
```
gluax, line 14: Then this should also have a call `handle` that consumes `self` and bubbles down.

gluax (changes requested):
Looking good so far! There are some things we can definitely clean up a little, though. And a few food for thought items relevant to the changes in this PR.

cli/src/cli/commands/node/get/node.rs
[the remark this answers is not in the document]
mariocao: In this case it is not directly the node config, because I still need to derive the public key from the Ed25519 keypair and encode it in hex. Anyhow, I think I managed to do it a little bit better.

cli/src/cli/commands/node/get/nodes.rs
```diff
@@ -19,15 +24,17 @@ impl Nodes {
     pub async fn handle(self, config: AppConfig, chains_config: PartialChainConfigs) -> Result<()> {
         let chains_config = config.chains.to_config(chains_config)?;
 
-        let contract_account_id = config.node.to_contract_account_id(self.contract_id)?;
+        let contract_id = if let Some(contract_id) = self.contract_id {
```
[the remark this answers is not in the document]
mariocao: Same as before

config/src/configs/delegate.rs
```diff
@@ -36,7 +36,7 @@ pub struct PartialDelegateConfig {
 #[cfg(feature = "delegate-cli")]
 impl PartialDelegateConfig {
     pub fn to_config(self, cli_options: Self) -> Result<DelegateConfig> {
-        let validator_secret_key = merge_config_cli!(self, cli_options, validator_secret_key, Ok(String::new()))?;
+        let validator_master_key = merge_config_cli!(self, cli_options, validator_secret_key, Ok(String::new()))?;
```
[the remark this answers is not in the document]
mariocao: That was not my code 😅

runtime/core/src/imports.rs
```diff
@@ -223,6 +223,24 @@ pub fn shared_memory_read_length_import_obj(store: &Store, vm_context: VmContext
     Function::new_native_with_env(store, vm_context, shared_memory_read_length)
 }
 
+/// Reads the value from memory as byte array and sends the number of bytes to
```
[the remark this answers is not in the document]
mariocao: Opps...it was copy-pasted.

wasm/consensus/src/tasks/batch.rs
mariocao: It was replaced not removed by `log!(Level::Debug, "[BatchTask] Starting task for contract id: {contract_id}");`

wasm/consensus/src/types/batch_signature.rs
mariocao: I changed it to a constant:

```
const EMPTY_SHA256: [u8; 32] = [
    227, 176, 196, 66, 152, 252, 28, 20, 154, 251, 244, 200, 153, 111, 185, 36, 39, 174, 65, 228, 100, 155, 147, 76,
    164, 149, 153, 27, 120, 82, 184, 85,
];
```

wasm/consensus/src/tasks/p2p.rs
mariocao: Agree, we can do it once we have more message types :)

cli/src/cli/commands/node/get/node.rs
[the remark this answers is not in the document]
gluax: I personally still think it's cleaner to have that logic in a function on the type. Also if we need similar logic in the future it's easy to just call that function then.

config/src/configs/delegate.rs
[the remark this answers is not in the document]
gluax: Oops my bad haha

wasm/consensus/src/tasks/batch.rs
gluax: But that's in the handle function. Not in batch step 1. That assumes we will always make it to batch step 1 which isn't necessarily true.

mariocao:
I just tried out the code and updated the testing instructions. 🤯

jamesondh (approved):
Working for me, but for future reference (during end-to-end testing) I had to set two config values manually to get it to work with my localnet script, when they should be using `SEDA_NEAR_RPC_URL` in env or `chain_rpc_url` in `config.toml`:

1. `RPC_URL` in `DelegateConfigInner`
2. `chain_rpc_url` in `test_config()`

If you want to test this on localnet, run the script `sh deploy-localnet.sh --committeeSize 2 --testAccounts true` on branch `feat/deploy-localnet` and follow the instructions to process the epoch and start the nodes once the setup is complete.

crypto/src/test.rs
Thomasvdam, line 1: Perhaps add a test to verify that changing the index produces a different key?

Thomasvdam (approved):

PR #115, opened 2023-04-05
Remove promises & migrate all binaries over to new wasmer 3

FranklinWaller:
## Motivation

Removes the usage of promises and migrate the existing WASM binaries over to sync calls. This greatly simplifies the binaries since there is no need any more for different steps. Also fixed a bug where the chain_call would get stuck if a transaction was invalid (if the node didn't have enough balance).

## Explanation of Changes

* Removed promises from the runtime (including the promise queue and handling of it)
* Replaced promises with sync function inside all WASM binaries
* Implemented cache for WASM binaries (with Wasmer 2 we had implemented that but it works differently now)
* Use `RpcBroadcastTxCommitRequest` instead of `RpcBroadcastTxAsyncRequest` for better visibility on why a transaction failed. (instead of it being gone forever)
* Implemented easy to use parse functions for call results inside WASM.
* Removed memory read/write since it's useless now (shared is still there)
* Fixed tests 

## Testing

The usual test command covers all the cases but the chain_call one.
The best way to test that one is running the normal consensus setup (having multiple nodes run and letting the leader send a batch)


common/src/node/get.rs
```diff
@@ -55,6 +55,12 @@ pub struct ComputeMerkleRootResult {
     pub current_slot_leader: Option<String>,
 }
 
+impl From<Vec<u8>> for ComputeMerkleRootResult {
```
gluax, line 58: Maybe using `TryFrom` instead so we don't `.unwrap` here.

config/src/configs/node.rs
```diff
@@ -35,6 +33,9 @@ pub struct PartialNodeConfig {
     /// An option to override the node runtime worker threads config value.
     #[arg(long)]
     pub runtime_worker_threads:  Option<u8>,
+    // An option to override the default path where to store the WASM cache
+    #[arg(long)]
+    pub wasm_cache_path:         Option<PathBuf>,
```
gluax, line 38: We can discuss this at a later point, but imo any and all caching should go to the standard app settings or cache directories as per the OS itself.

For the default location that is. We can still allow them to override it if they really desire.

runtime/sdk/src/promises/promise.rs
```diff
@@ -22,6 +22,24 @@ pub enum PromiseStatus {
...
 
+impl PromiseStatus {
+    /// Helper function that immidiatly assumes that the promise has been
+    /// fulfilled and return the value. Panics if result is not fulfilled
+    pub fn fulfilled(self) -> Vec<u8> {
+        if let Self::Fulfilled(Some(value)) = self {
+            return value;
+        }
+
+        panic!("Promise is not fulfilled: {:?}", &self);
```
gluax, line 33: I think it's fine for now since we haven't been handling errors much in our wasm binaries yet and panicking whenever we error.

But we should eventually do an error hanging there as well for better error message experience for users.

gluax (approved):
Amazing man!!!

Just had one comment on a change from using `From` to `TryFrom`. The others are just future things to note/discuss.

runtime/sdk/src/promises/promise.rs
FranklinWaller: Yeah panics are not destructive in WASM so we are fine in that regard but i do agree that we need to do some error handling (I like using log!(Level::Error) for this case)

config/src/configs/node.rs
FranklinWaller: Didn't give it really a second thought, I personally kinda like everything being visible what the app stores

PR #116, opened 2023-04-05
feat: use uncompressed pks and sigs

mennatabuelnaga:
## Motivation
To avoid back and forth conversions in `bn254_verify()`'s `format_pairing_check_values()`, and to reduce consequent gas costs in `post_signed_batch()`

## Explanation of Changes

Refactors everything to use uncompressed signatures and uncompressed public keys.

## Related PRs and Issues
https://github.com/sedaprotocol/bn254/pull/7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,10 +48,7 @@ actix = "0.13"
 async-trait = "0.1"
 bs58 = "0.4.0"
 base64 = "0.13"
-
-# bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
-bn254 = { git = "https://github.com/mennatabuelnaga/bn254", branch = "main" }
-
+bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
 borsh = { version = "0.9", default-features = false }
 clap = { version = "4.1", default-features = false }
 clap-markdown = { version = "0.1", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,10 @@ actix = "0.13"
 async-trait = "0.1"
 bs58 = "0.4.0"
 base64 = "0.13"
-bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
+
+# bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
+bn254 = { git = "https://github.com/mennatabuelnaga/bn254", branch = "main" }
+
 borsh = { version = "0.9", default-features = false }
 clap = { version = "4.1", default-features = false }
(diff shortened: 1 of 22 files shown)


common/src/node/mod.rs
```diff
@@ -41,7 +41,12 @@ impl NodeInfo {
                 rng.gen_range(1..65535)
             ),
             balance:            rng.gen_range(0..u128::MAX),
-            bn254_public_key:   master_key.derive_bn254(0).unwrap().public_key.to_compressed().unwrap(),
+            bn254_public_key:   master_key
```
gluax, line 44: It seems like we are using the uncompressed version a lot more.

I wonder if it's better to store the uncompressed versions then rather than having to call `.to_uncompressed` when we access it?

gluax (commented):
Code looks fine to me, but I don` t like how we call `to_uncompressed` in so many locations where it seems we might be able to already have it done. Not sure tbh, so opening discussion here to see if anyone has a good idea :)

Cargo.toml
```diff
@@ -48,7 +48,10 @@ actix = "0.13"
 async-trait = "0.1"
 bs58 = "0.4.0"
 base64 = "0.13"
-bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
+
+# bn254 = { git = "https://github.com/sedaprotocol/bn254", branch = "main" }
+bn254 = { git = "https://github.com/mennatabuelnaga/bn254", branch = "main" }
```
FranklinWaller, line 53: We should only merge this PR once this is using the version on sedaprotocol again

common/src/node/mod.rs
FranklinWaller: If it's possible i would also go for this option. I think in the long run it's more gas to convert then to just store the larger key

FranklinWaller: From the node's side it's ok but for the contract i would not do conversions

FranklinWaller: @mennatabuelnaga can the node help by sending always the uncompressed version?

FranklinWaller: Nevermind, most of it is just tests, it's fine in my book :)

FranklinWaller (approved):

common/src/node/mod.rs
mennatabuelnaga: I think keeping it as is may be more consistent
i.e: In crypto/keypair.rs:
 pub type Bn254KeyPair = KeyPair<Bn254PrivateKey, Bn254PublicKey>;
is more consistent than
pub type Bn254KeyPair = KeyPair<Bn254PrivateKey, Vec<u8>>;

jamesondh (approved):

PR #118, opened 2023-04-06
feat(cli): Add seda init command

FranklinWaller:
## Motivation

Currently the flow is kinda full automatic where doing `seda run` generates a bunch of stuff and then runs immediately. Most cases the user is not done configuring. In order to more stream line the node operator experience this introduces a simple `seda init` command that generates a secret key and config.toml and then displays the public keys that was derived from that secret key.

## Explanation of Changes

* Add new `seda init` command

## Testing

Run `seda init`

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -9,13 +9,21 @@ name = "seda"
 path = "src/main.rs"
 
 [dependencies]
+async-trait = { workspace = true }
 bn254 = { workspace = true }
+bs58 = { workspace = true }
 clap = { workspace = true, features = ["default"] }
-clap_complete = { workspace = true }
 clap-markdown = { workspace = true }
+clap_complete = { workspace = true }
 dotenv = { workspace = true }
+ed25519-dalek = { workspace = true }
 hex = { workspace = true }
 jsonrpsee = { workspace = true, features = ["client"] }
+near-crypto = { workspace = true }
+near-jsonrpc-client = { workspace = true }
+near-jsonrpc-primitives = { workspace = true }
+near-primitives = { workspace = true }
+rand = "0.7"
 seda-chains = { workspace = true }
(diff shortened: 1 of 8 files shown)


gluax (approved):

cli/Cargo.toml
Thomasvdam, line 1: I'm sure there's a good reason for this, but I don't see all of these dependencies within the changes of this PR? What am I missing?

Thomasvdam (approved):

cli/Cargo.toml
gluax: it looks to me like a rebase/merge, I think a lot of them actually aren't needed since they were removed on main.

cli/src/cli/mod.rs
```diff
@@ -53,6 +53,7 @@ pub enum Command {
         #[command(subcommand)]
         sub_chain_command: SubChain,
     },
+    Init(Init),
```
mariocao, line 56: Missing doc (to be displayed when calling `--help`)

cli/src/cli/commands/init.rs
```diff
@@ -0,0 +1,57 @@
...
+        create_and_load_or_load_config(Some(PathBuf::from(self.config_path.to_string())));
+
+        println!("Key information: \n");
+        println!("NEAR Account ID: {account_id}");
+        println!(
+            "ED25519 Public Key (base58 encoded): ed25519:{}",
+            bs58::encode(&ed25519_public_key).into_string()
+        );
+        println!(
+            "BN254 Public Key (base58 encoded): {}",
```
mariocao, line 51: Should we display the `bn254` public key with the same format as the `ed25519`?

```suggestion
            "BN254 Public Key (base58 encoded): bn254:{}",
```

mariocao (approved):

PR #120, opened 2023-04-13
[Feat] Deploy Multiple Nodes

gluax:
## Motivation

To have a singular command to launch multiple nodes locally easily.

## Explanation of Changes

Had to make some changes to the output of the `seda init` command to make it easier to machine parse. This change also benefits users.

Otherwise, it adds a new command to the debugger: 
`make run-seda-debug deploy-multiple`.

This defaults to running 3 nodes.
You can add `-a 5` to run 5 nodes.
These nodes save their data to `./seda_dbg_files/saved_nodes.json`.
Their logs are saved to `./seda_dbg_files/node_logs/node_#_log...`
Their sks are saved to `./seda_dbg_files/node_sks/node_#_sk`.

If you specify a number less than or equal to the number of saved nodes, it will use the already saved info without generating more.
Specifying a number greater than the number of saved nodes will generate the number of missing nodes.

The organization and setup of this functionality were also set up with fuzzing in mind since a lot of this behavior will need to be re-used.

## Testing

Have test this since it's just running the node manually.

Here's the folder structure:
image

Here's what the JSON file looks like:
image

Here's two of the node logs so you can see they are all connecting:
image

image

## Related PRs and Issues

None


CLI.md
```diff
@@ -411,3 +397,5 @@ Views the specified method on the specified chain with the given args and contra
     This document was generated automatically by
     <a href="https://crates.io/crates/clap-markdown"><code>clap-markdown</code></a>.
 </i></small>
+ href="https://crates.io/crates/clap-markdown"><code>clap-markdown</code></a>.
+</i></small>
```
Thomasvdam, line 401: I think the tool had a small hiccup here.

tooling/debug/src/common/port_info.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// Generates a random valid port number and checks if it was seen first, and
+/// afterwards checks if it's in use. Keeps doing this till it get's a valid
+/// port.
+fn generate_free_port<R: Rng + ?Sized>(seen_ports: &mut HashSet<u16>, rng: &mut R) -> u16 {
+    let mut port = rng.gen_range(5000..=65535);
+
+    while !seen_ports.contains(&port) && !is_port_free(port) {
+        seen_ports.insert(port);
+        port = rng.gen_range(1..=65535);
```
Thomasvdam, line 30: Why is this range different from the one above?

Thomasvdam (commented):
Do we need to consider that previously used ports are not guaranteed to be available on a next invocation?

Some of the rusty stuff still goes over my head so leaving it to more experienced team members to 'pass judgement' :P

CLI.md
gluax: Good catch! I'll fix that.

tooling/debug/src/common/port_info.rs
gluax: Oops, I was gonna start our port generation from 5000, so we are less likely to hit conflicting apps but forgot to change it in both places :). I'll change that to a global const.

gluax:
@Thomasvdam it's true I don't recheck the port after loading it from a file. I figured if something does clash we can just edit the port in the file. Since this tool is mostly for devs.

But if you want I can add some logic for rechecking?

Thomasvdam:
I think if it errors with a message detailing which node/port is the problem we're good 👍 Seems too much of an edge case to add additional logic for.

gluax:
@Thomasvdam I just double-checked; it doesn't quit the program, unfortunately. But this is due to the `node` not having exit codes. So even when it errors, it exits with a `success` instead of a `failure`.

As a temporary workaround, I'm allowing stderr to leak to the console so you can `ctrl-c` after realizing the error:
image

Then I'll make a backlog issue to handle exit codes in the `node`, and we can eventually remove the stderr printing to stdout.

mariocao (approved):

